PR #56, opened 2015-01-12
Modify popover function for showing "Test Log"

GwonGisoo:
- Change the toggle event to click
- Apply scroll-y in popover

### DIFF
--- a/ngrinder-controller/src/main/webapp/WEB-INF/ftl/perftest/detail.ftl
+++ b/ngrinder-controller/src/main/webapp/WEB-INF/ftl/perftest/detail.ftl
@@ -9,6 +9,11 @@
 	<link href="${req.getContextPath()}/css/slider.css" rel="stylesheet">
 	<link href="${req.getContextPath()}/plugins/datepicker/css/datepicker.css" rel="stylesheet">
 	<style>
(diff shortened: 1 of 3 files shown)


newpcraft:
Thanks for your code, but, it has some problems.

Although it provided new method to show a result by clicking, maybe new user couldn't know how to see the result. So, I'd like to add a title tag to show a hint if clicking is the best way to provide a result.

Another issue, it maybe an important issue, it would be globally affected, so, on some UI, it doesn't work what you wanted.

I think, it would be implemented as following:

- support mouse over to see the result (as same as before)
- when focus inside the result pane, it should be kept display
- if mouse out of the result pane, it should be disappeared (or if it is difficult to implement. you might possible to handle click event and if click event received out of result pane area, it try to close the displayed pane)
- focusing area should be expanded, which means that it should be affected to focus on the square area out of bullet.
skitched-20150114-121007

Would you please check above issue?

GwonGisoo:
thanks feedback. I will try that!

GwonGisoo:
커밋 로그가 꼬여서 이 pull-request는 제거하고 다시 진행하겠습니다..ㅜㅜ

PR #57, opened 2015-01-15
Modify popover function for showing "Test Log"

GwonGisoo:
- Add function to disable hide at "click event" of popover
- Apply scroll-y in popover


newpcraft:
If the user clicked at the step 1-1, and then click another ball element of test, does it hide the previous popup dialog?

GwonGisoo:
Yes, if the user clicks on the other ball, hide in less than one second.

PR #74, opened 2015-08-13
Modiried agent file download logic

nanpa83:
Dear all, 

I have modified agent scripts of both standalone and multi.
- to check whether an agent file exists or not before trying to download it from its controller. 
- only if the agent file does not exist, it is downloaded from the controller.

Please kindly check and accept it.
Thank you.


nanpa83:
I missed some features.

PR #75, opened 2015-08-13
modified agent binary download logic

nanpa83:
I have enhanced agent starting performance.
- deleting pid before starting the agent
- downloading the agent binary from the controller only when it doesn't exist. 

It supports "docker start" and "docker stop" scheme.


junoyoon:
Merged. It's reflected to ngrinder/agent:3.3-p1 and ngrinder/multi-agent:3.3-1 in docker hub.

Issue #109, opened 2016-04-19
Setting a TravisCI


GwonGisoo:

junoyoon:
@GwonGisoo  TravisCI seems not working yet. Could u share the progress on this?

GwonGisoo:
TravisCI is required to '.travis.yml' file on Project root path.
You can see result of TravisCI below:
https://travis-ci.org/naver/ngrinder/builds/124123067

### Travis run progress
1. TravisCI is detect to the new PR or Commit.
2. If PR or Commit was added, create the build environment to based on 'ROOT/.travis.yml'
3. When build environment is created, will be execute maven build by 'mvn install -DskipTests=true'.
4. When maven build is completed, will be execute maven test by 'mvn test'.

※ '.travis.yml' should be create a test evironment for jdk7 and jdk8.

GwonGisoo:
But now occur failed on 'mvn test'
```
testNotRegisteredDNSLookup(org.ngrinder.dns.LocalManagedDnsTest)  Time elapsed: 0.011 sec  <<< FAILURE!
```

junoyoon:
We need to skip the test.. Just compile.

junoyoon:
Fixed

PR #110, opened 2016-04-19
#109 Setting a TravisCI

GwonGisoo:
Add a travis yml file


GwonGisoo:
This commit is do test on jdk7, jdk8, but test failed yet.

Issue #103, opened 2016-04-14
Fix security issue


songeunwoo:
Fix ngrinder security issue.

junoyoon:
Fixed

songeunwoo:
In case of using generic type, it turns into an unnecessary filter so that it needs to be revised when creating a stress script.

PR #121, opened 2016-04-26
#103 Fix security issue

songeunwoo:
- Luxy XSS filter is applied to the script validation and save unintentionally.


junoyoon:
Could you make the commit comment clear? I can not understand what it is. @songeunwoo

songeunwoo:
When creating a script file that uses the generic type,
Need to modify the generated issues.

Issue #134, opened 2016-05-03
Support CJK for folder and script name


chengaomin:
I find the url is wrong when I saved my script.
(my url contains Chinese)

`http://127.0.0.1:9000/script/detail/测试/pro/test.py`  to  `http://127.0.0.1:9000/script/list/??/pro/`

![](http://7xrn7f.com1.z0.glb.clouddn.com/16-5-3/65894915.jpg)

then

![](http://7xrn7f.com1.z0.glb.clouddn.com/16-5-3/82849542.jpg)

Can it support Chinese?

junoyoon:
I have to know.. your env.
Did u run ngrinder on the tomcat or just run it by java -jar ngrinder.war way?

chengaomin:
@junoyoon by java -jar

junoyoon:
We found this happens when running ngrinder using embedded webserver or running it on old apache tomcat.

We can not fix this issue by 3.4.0 but will fix it by 3.4.1. Please run ngrinder on tomcat over 7.0.68 by then.

junoyoon:
I've tried fix this just before. It's easily fixed than I though.

```
	return "redirect:/script/list/" + encodePathWithUTF8(basePath);
```

junoyoon:
Fixed.

PR #138, opened 2016-05-04
#134 Chinese in url

junoyoon:
@songeunwoo  Please review this.


songeunwoo:
It is a good solution.
no problem testing in progress.

songeunwoo:
It is without error.
no more than. good~

Issue #177, opened 2016-06-12
agent_monitor.conf配置文件描述有误


aizaimenghuangu:
ngrinder/ngrinder-controller/src/main/resources/ngrinder_agent_home_template/agent_monitor.conf
中描述的**monitor.binding_port=hostname_or_ip** 有误，修正为：monitor.binding_**ip**=hostname_or_ip

参考来源：
ngrinder/ngrinder-core/src/main/java/org/ngrinder/common/constants/MonitorConstants.java:
public interface MonitorConstants {
	**public static final String PROP_MONITOR_BINDING_IP = "monitor.binding_ip";**
	public static final String PROP_MONITOR_BINDING_PORT = "monitor.binding_port";

	/**
	 * Monitor Constant
	 */
	public static final int DEFAULT_MONITOR_COLLECTOR_INTERVAL = 1;
	public static final String DEFAULT_MONITOR_DOMAIN = "org.ngrinder.monitor";
	public static final String SYSTEM = "name=System";
	public static final String MONITOR_FILE_PREFIX = "monitor_system_";
}

junoyoon:
FIXED. Thanks

PR #178, opened 2016-06-13
 #177 modify agent_monitor.conf

songeunwoo:
modify agent_monitor.conf


junoyoon:
Please split commits.

PR #190, opened 2016-07-19
modify convert escape

songeunwoo:
modify editor.ftl


junoyoon:
I can not sure this is appropriate. I'll close the PR.

PR #215, opened 2016-10-17
update install_maven_lib.bat

lunker:
In Windows, cmd, powershell, or git bash can't found mvnw script in ngrinder folder.
so add current path to distinguish right mvnw.


junoyoon:
I don't understand why this is necessary. mvnw is located in the same location with install_maven_lib.sh  so it's callable well.

junoyoon:
close this because there are no more discussion.

PR #235, opened 2017-03-18
Fix csv separator inside SingleConsole

tolis-e:
The hardcoded comma separator has been replaced with the dynamic one.


junoyoon:
fixed by #295

Issue #273, opened 2018-01-05
interval is not used in getMonitorGraph method in class PerfTestService


hongzenghuang:
during the test, i find that the label of xAxis in tps graph and cpu/memory graph are different. while checking the code, find that in the getMonitoGraph method, the interval is not used which result in that this method return all monitor values to UI, then the cpu/memory xAxis value is incorrect.

May change to below:
while (StringUtils.isNotBlank(line)) {
				if (skipCount < dataInterval) {
					skipCount++;
                                        **line = br.readLine();**
				} else {
					skipCount = **0**;
					String[] datalist = StringUtils.split(line, ",");
					if ("null".equals(datalist[4]) || "undefined".equals(datalist[4])) {
						sbUsedMem.append("null").append(",");
					} else {
						sbUsedMem.append(Long.valueOf(datalist[4]) - Long.valueOf(datalist[3])).append(",");
					}
					addCustomData(sbCPUUsed, 5, datalist);
					addCustomData(sbNetReceived, 6, datalist);
					addCustomData(sbNetSent, 7, datalist);
					addCustomData(customData1, 8, datalist);
					addCustomData(customData2, 9, datalist);
					addCustomData(customData3, 10, datalist);
					addCustomData(customData4, 11, datalist);
					addCustomData(customData5, 12, datalist);
					
				}
			}

hongzenghuang:
the same issue with getReportPluginGraph method

PR #274, opened 2018-01-12
#273 fix interval is not used in getMonitorGraph and getReportPluginG…

hongzenghuang:
…raph method in class PerfTestService


junoyoon:
Thx for this fix.

PR #278, opened 2018-02-26
fix bug when har file has no params

raoweijian:
image
the params may be null, and this will cause the null pointer.


junoyoon:
Thanks for the PR.

PR #279, opened 2018-02-28
add more stats in detail report

keithmork:
Add error rate, percentiles, min RT, max RT in detail report.

English UI

Chinese UI


junoyoon:
Sorry for late response.
Thanks for PR. 
It seems little complicated. isn't it better to make it compact.

lijie1010:
I have a good idea to make it compact.

imbyungjun:
Thank you for your PR. We will positively consider your proposal in the next version.

PR #286, opened 2018-05-13
The name of param in javadoc does not match which in method

zhaogp:
Maybe change it in method but not in javadoc


junoyoon:
Thx for this push. 
We're rewriting the ngrinder right now.  So this is not so worthwile to merge this.
However, we will keep your comment in mind.

PR #293, opened 2018-07-23
Add system configuration for setting security level

donggyu04:
## Distinguish nGrinder's security level
To distinguish nGrinder's security level I add NGrinderLightSecurityManager which less secured than NGrinderSecurityManager.

This allows followings.(not allowed in NGrinderSecurityManager.)

**1. multicast
2. tcp connection unspecified address**

If you want to use this, set 'controller.security.level=light'. It only works if controller.security is true.
 
The default value for this configuration is 'controller.security.level=normal'.
(using NGrinderSecurityManager.)

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/common/constant/ControllerConstants.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/common/constant/ControllerConstants.java
@@ -68,6 +68,7 @@ public interface ControllerConstants {
 	public static final String PROP_CONTROLLER_SAFE_DIST = "controller.safe_dist";
 	public static final String PROP_CONTROLLER_SAFE_DIST_THRESHOLD = "controller.safe_dist_threshold";
 	public static final String PROP_CONTROLLER_SECURITY = "controller.security";
+	public static final String PROP_CONTROLLER_SECURITY_LEVEL = "controller.security.level";
 	public static final String PROP_CONTROLLER_URL = "controller.url";
 	public static final String PROP_CONTROLLER_VALIDATION_SYNTAX_CHECK = "controller.validation_syntax_check";
 	public static final String PROP_CONTROLLER_USAGE_REPORT = "controller.usage_report";
--- a/ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
@@ -57,6 +57,7 @@ import java.util.Properties;
 
 import static net.grinder.util.NoOp.noOp;
 import static org.ngrinder.common.constant.DatabaseConstants.PROP_DATABASE_UNIT_TEST;
+import static org.ngrinder.common.constants.GrinderConstants.GRINDER_SECURITY_LEVEL_NORMAL;
 import static org.ngrinder.common.util.Preconditions.checkNotNull;
 
 /**
@@ -551,6 +552,15 @@ public class Config extends AbstractConfig implements ControllerConstants, Clust
 	 */
 	public boolean isSecurityEnabled() {
 		return !isDevMode() && getControllerProperties().getPropertyBoolean(PROP_CONTROLLER_SECURITY);
+	}
+
+	/**
+	 * Get system security level from system properties.
+	 *
+	 * @return security level.
+	 */
+	public String getSecurityLevel() {
+		return getControllerProperties().getProperty(PROP_CONTROLLER_SECURITY_LEVEL, GRINDER_SECURITY_LEVEL_NORMAL);
 	}
 
 	/**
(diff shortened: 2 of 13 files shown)


donggyu04:
I modified NGrinderSecurityManager to prevent access to nGrinder home folder from test script code.

Users only access their work folder.

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -50,9 +56,25 @@
 	}
 
 	void init() {
+		controllerHomeDir = resolveControllerHomeDir();
```
junoyoon: Why does it resolve controller home? is it better to name it with HomeDir?  It's because this class not only called from controller but agent.

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -191,16 +213,16 @@ public void checkPermission(Permission permission, Object context) {
 
 	@Override
 	public void checkRead(String file) {
-		if (file != null && file.contains("database.conf")) {
-			throw new SecurityException("File Read access on database.conf is not allowed.");
+		if (file != null) {
+			this.nGrinderHomeReadAllowed(file);
```
junoyoon: The method name is weird.

this.isReadAllowed might be better.

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
junoyoon: However, this code can be executed agent machine.. and mistakenly configured env var can cause the error.
I think it might be better to provide the way to detect ngrinder controller context.
Can we inherit PropertyBuilder (named ValidationPropertyBuilder) and add the additional param in buildJVMArgumentWithoutMemory method and initiate it in the LocalScriptTestDriveService ?

donggyu04: Yes, I understand.

I will change the read access logic in security manager according to the jvm argument "ngrinder.context".

junoyoon: Thx

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -57,6 +58,7 @@
 
 	void init() {
 		controllerHomeDir = resolveControllerHomeDir();
+		ngrinderContext = System.getProperty("ngrinder.context", "agent");
```
junoyoon: is it better to use boolean type?

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -217,7 +217,7 @@ public void checkPermission(Permission permission, Object context) {
 	public void checkRead(String file) {
 		if (ngrinderContext.equalsIgnoreCase("controller")) {
 			if (file != null) {
-				this.nGrinderHomeReadAllowed(file);
+				this.isReadAllowed(file);
```
junoyoon: Why is this located in the NGrinderSecurityManager.java not Light One?

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
junoyoon: Still is it better to resolve ngrinder home only in controller context?

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -191,16 +215,20 @@ public void checkPermission(Permission permission, Object context) {
 
 	@Override
 	public void checkRead(String file) {
-		if (file != null && file.contains("database.conf")) {
-			throw new SecurityException("File Read access on database.conf is not allowed.");
+		if (ngrinderContext.equalsIgnoreCase("controller")) {
```
junoyoon: Is it better to make isControllerContext method?

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -223,6 +251,20 @@ public void checkExec(String cmd) {
 		throw new SecurityException("Cmd execution of " + cmd + " is not allowed.");
 	}
 
+	/**
+	 * File read access is allowed on only user-work and tmp directory when access ngrinder home folder
```
junoyoon: It's better to make this comment like following.

```
/**
*  Check if the given file is safe to read.
*/
```

junoyoon:
LGTM

PR #296, opened 2018-07-30
fix typo error

donggyu04:
referred to #245


junoyoon:
It's wrong fix. I'll revert it.

PR #299, opened 2018-08-03
Modify NGrinderSecurityManager read access 

donggyu04:
## Modify NGrinderSecurityManager read access
I modified NGrinderSecurityManager read access. It will work like this.

1. ngrinder controller
   - Restrict read access for ngrinder home. but user can access their work and tmp folder in there
2. ngrinder agent
   - No restrict


junoyoon:
LGTM

PR #309, opened 2018-08-13
ignore non-test target class

donggyu04:
Ignore non-test target class


junoyoon:
This PR doesn't follow the git flow naming scheme.
I'll create new PR.

PR #318, opened 2018-08-16
Bug fix from lack of required attributes in login.ftl

donggyu04:
When i login system user account, I encountered this error at `login.ftl`

image

So i added required attributes before login view was resolved.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/home/controller/HomeController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/home/controller/HomeController.java
@@ -128,6 +128,7 @@ public class HomeController extends BaseController implements ControllerConstant
 				// set local language
 				setLanguage(getCurrentUser().getUserLanguage(), response, request);
 				setLoginPageDate(model);
+				setLoginRequiredAttributes(model);
 				role = user.getRole();
 			} catch (AuthenticationCredentialsNotFoundException e) {
 				return "login";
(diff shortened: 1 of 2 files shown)


ngrinder-controller/src/main/java/org/ngrinder/home/controller/HomeController.java
```diff
@@ -266,6 +265,12 @@ public String login(ModelMap model) {
 		return "redirect:/";
 	}
 
+	private void setLoginRequiredAttributes(ModelMap model) {
```
junoyoon: what about changing the method name.. like..

addCommonLoginAttribute

ngrinder-controller/src/main/java/org/ngrinder/home/controller/HomeController.java
```diff
@@ -266,6 +265,12 @@ public String login(ModelMap model) {
...
+		model.addAttribute("signUpEnabled", getConfig().isSignUpEnabled());
+		final String defaultLang = getConfig().getControllerProperties().getProperty(ControllerConstants.PROP_CONTROLLER_DEFAULT_LANG);
+		model.addAttribute("defaultLang", defaultLang);
+	}
+
 	private void setLoginPageDate(ModelMap model) {
```
junoyoon: What about changing the method name like..
addTimezoneAttribute

PR #319, opened 2018-08-17
Remove unnessesary test

donggyu04:
I removed test `ClusteredAgentManagerServiceTest.testOther` because the test is meaningless and affects other test.


junoyoon:
LGTM

PR #329, opened 2018-08-21
Pass working directory to test process to avoid wrong ngrinder home path on symbolic lynk

donggyu04:
Java `System.getProperty ("user.dir")` return the link value when current execution location is a symbolic link. so `workDirectory` value in security manager is different from what i expected.

This commit passes ```ngrinder.base.path``` to test process as a system property. So the path resolution  in security manager can handle the file access well.

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
+++ b/ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
@@ -176,6 +176,7 @@ public class PropertyBuilder {
 		jvmArguments = addParam(jvmArguments, properties.getProperty("grinder.param", ""));
 		jvmArguments = addPythonPathJvmArgument(jvmArguments);
 		jvmArguments = addCustomDns(jvmArguments);
+		jvmArguments = addUserDir(jvmArguments);
 		if (server) {
 			jvmArguments = addServerMode(jvmArguments);
 		}
@@ -365,6 +366,11 @@ public class PropertyBuilder {
 		if (enableLocalDNS) {
 			jvmArguments.append(" -Dsun.net.spi.nameservice.provider.1=dns,LocalManagedDns ");
 		}
+		return jvmArguments;
+	}
+
+	private StringBuilder addUserDir(StringBuilder jvmArguments) {
+		jvmArguments.append(" -Duser.dir=" + baseDirectory.getFile().getPath() + " ");
 		return jvmArguments;
 	}
 


junoyoon:
LGTM

ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
```diff
@@ -176,6 +176,7 @@ public String buildJVMArgumentWithoutMemory() {
 		jvmArguments = addParam(jvmArguments, properties.getProperty("grinder.param", ""));
 		jvmArguments = addPythonPathJvmArgument(jvmArguments);
 		jvmArguments = addCustomDns(jvmArguments);
+		jvmArguments = addBasePath(jvmArguments);
```
junoyoon: I think this can be located in the line 171 so that it can be only activated in security mode.
And I think the more appropriate method name is... addUserDir

ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
```diff
@@ -368,6 +369,11 @@ private StringBuilder addCustomDns(StringBuilder jvmArguments) {
 		return jvmArguments;
 	}
 
+	private StringBuilder addBasePath(StringBuilder jvmArguments) {
+		jvmArguments.append(" -Dngrinder.base.path=" + baseDirectory.getFile().getPath() + " ");
```
junoyoon: Isn't it better to use ngrinder.user.dir ?

ngrinder-runtime/src/main/java/org/ngrinder/sm/NGrinderSecurityManager.java
```diff
@@ -55,6 +55,7 @@
 
 	void init() {
 		ngrinderContext = System.getProperty("ngrinder.context", "agent");
+		workDirectory = System.getProperty("ngrinder.base.path", System.getProperty("user.dir"));
```
junoyoon: Isn't it better to use ngrinder.user.dir ?

donggyu04:
@junoyoon 
Thank you for your review. I applied that.

junoyoon:
I tested like below.

```
Naverui-MacBook-Pro-53:.ngrinder Naver$ java -Duser.dir=/Users/Naver/.ngrinder A
/Users/Naver/.ngrinder
```

I found that we can pass user.dir
So I think it's better to pass user.dir not ngrinder.user.dir

If so.. we can revert the changes on NGrinderSecurityManager.java

junoyoon:
And I think we can apply user.dir parameter always (not only security enabled) if we decide to use this.
I think it will enhance the script executability(?)

junoyoon:
build this

junoyoon:
build this

PR #330, opened 2018-08-22
Provide the way to turn off security mode in agent config.

donggyu04:
Add `agent.enable_security` configuration to `agent_conf`.

It is used to operate regardless of the `controller.security` configuration of the controller.

when `agent.enable_security=false` the agent always runs in unsecurity mode.

default value is true.(It depends on controller security configuration.)


junoyoon:
LGTM

PR #333, opened 2018-08-24
Prevent agent list flikering when initial loading

donggyu04:
I prevent agent list from flickering.

It seems to be due to the difference between when the list is shown and when the paging is applied.

so I show agent list in `fnInitComplete` callback.(In `fnInitComplete`, paging has already been applied.)


junoyoon:
LGTM

Issue #325, opened 2018-08-18
Move to Gradle


junoyoon:
Maven which is used in nGrinder does not fit smoothly.

1. We need to run install-maven-lib to avoid missing dependency.
2. We need to update all versions in sub modules whenever releasing new version.
3. inter-module dependency is not handled if it's not explicitly built 

I think it's time to move to Gradle.

junoyoon:
Merged.

PR #339, opened 2018-09-12
Move to Gradle

donggyu04:
#325

Converted build tool fronm Maven to Gradle.
Test completed in Mac or Windows environment, It looks like working well.
Some tests were failed in Gradle environment. I temporarily ignored it, I will deal with later.


build.gradle
```diff
@@ -0,0 +1,53 @@
+allprojects  {
+  apply plugin: "maven"
```
junoyoon: why do we need maven here?

ngrinder-core/build.gradle
```diff
@@ -0,0 +1,64 @@
...
+    }
+    compile (group: 'dnsjava', name: 'dnsjava', version:'2.1.1') {
+        exclude (module: 'junit')
+    }
+}
+
+task makeAgentDependenciesFile {
+    doFirst {
+        def classpath = ""
+        Pattern p = Pattern.compile("(.*?)-\\d+")
```
junoyoon: regular expression in groovy can be simpler than this.

check http://www.tothenew.com/blog/regex-in-groovy/

ngrinder-groovy/build.gradle
```diff
@@ -0,0 +1,22 @@
+description = "ngrinder-groovy"
+dependencies {
+    compile (group: "org.json", name: "json", version:"20090211")
+    compile (group: "org.codehaus.groovy", name: "groovy-all", version:"2.2.1") {
+        force = true
```
junoyoon: why do we need force here?

junoyoon:
LGTM

PR #344, opened 2018-10-01
Add defence code for NPE when get ready agent count

donggyu04:
When agent configuration `region` is empty such as `region=` NPE occurs in `getReadyAgentCount`.

Therefore, use `StringUtils.equals` to compare strings.


junoyoon:
LGTM

PR #346, opened 2018-10-24
Add character encoding option in mysql connection string

donggyu04:
When user selects mysql as their database, Automatically set character encoding to utf-8.


ngrinder-controller/src/main/resources/ngrinder_home_template/database.conf
```diff
@@ -7,7 +7,7 @@ database.type=H2
 
 # if you apply additional options for mysql database enable this.
 # this is added after the 'database.url' config.
-#database.url_option=?characterEncoding=utf8
+#database.url_option=&
```
junoyoon: Isn't it better to put & in the url template?

junoyoon:
LGTM

Issue #357, opened 2018-12-20
new version error


keny16888:
APPLICATION FAILED TO START
***************************

Description:

There is a circular dependency between 3 beans in the application context:
        - DBInit (field private org.springframework.security.authentication.dao.SaltSource org.ngrinder.infra.init.DBInit.saltSource)
        - securityConfig (field private org.ngrinder.security.NGrinderAuthenticationProvider org.ngrinder.infra.config.SecurityConfig.nGrinderAuthenticationProvider)
        - ngrinderAuthenticationProvider (field private org.springframework.security.authentication.dao.SaltSource org.ngrinder.security.NGrinderAuthenticationProvider.saltSource)
        - reflectionSaltSource

donggyu04:
What is new version? 3.4.2 release? or latest develop branch?

donggyu04:
Could you tell me your local environment for running the build? (OS, jdk version, build tool version, etc..)

donggyu04:
@keny16888 

I confirmed this occur at mac os with `develop` branch. i'll fix that.

could you use `master` branch instead of `develop`?

PR #358, opened 2018-12-20
Use lazy loading for resolve circular dependency issue

donggyu04:
#357

Use lazy loading for resolve circular dependency issue.

I confirmed this in windows and mac.


junoyoon:
LGTM

PR #368, opened 2019-01-11
fixed  initialize myql failed

lijie1010:


donggyu04:

junoyoon:
@lijie1010  Could u give us the detailed description? why uis the initialization error  occurred?

lijie1010:
Maybe  mysql‘s version lead to this problem. my version is 5.7，
I have solved this problem by reference this 
[http://www.cnblogs.com/wang666/p/9186559.html]

donggyu04:
It doesn't seem to be a problem.
If there is any problem, I'll fix it later.

PR #371, opened 2019-01-25
Fixe not detected TooManyError.

donggyu04:
While apply to hazelcast, I have missed some changes.

1. modify `hasTooManyError` function to use hazelcast dist map.
2. clean up sampling and monitoring data from hazelcast after each test finished.
3. drop unused column from perftest table.
4. clean up unnecessary code.


junoyoon:
LGTM

PR #373, opened 2019-02-11
Fixed bug where parent fields could not be merged before saving.

donggyu04:
1. Add last modified date and user when init user table.
2. Fixed bug where parent fields could not be merged before saving.
     - `getClass().getDeclaredFields()` could not resolve parent fields.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/infra/init/DBInit.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/infra/init/DBInit.java
@@ -100,7 +100,13 @@ public class DBInit {
 			user.setRole(role);
 			user.setUserName(userName);
 			user.setEmail(email);
-			user.setCreatedDate(new Date());
+
+			Date now = new Date();
+			user.setCreatedDate(now);
+			user.setLastModifiedDate(now);
+			user.setCreatedUser(user);
+			user.setLastModifiedUser(user);
+
 			user = userRepository.save(user);
 			fileEntryService.prepare(user);
 		}
--- a/ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
+++ b/ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
@@ -19,6 +19,11 @@ import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import static org.ngrinder.common.util.NoOp.noOp;
 import static org.ngrinder.common.util.Preconditions.checkArgument;
@@ -33,6 +38,7 @@ import static org.ngrinder.common.util.Preconditions.checkNotNull;
 public abstract class ReflectionUtils {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(ReflectionUtils.class);
+	private static final Map<Class, List<Field>> fieldCache = new ConcurrentHashMap<>();
 
 	/**
 	 * Get object field value, bypassing getter method.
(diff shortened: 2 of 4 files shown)


ngrinder-controller/src/main/java/org/ngrinder/infra/init/DBInit.java
```diff
@@ -100,7 +100,12 @@ private void createUser(String userId, String password, Role role, String userNa
 			user.setRole(role);
 			user.setUserName(userName);
 			user.setEmail(email);
-			user.setCreatedDate(new Date());
+
+			Date createdDate = new Date();
```
junoyoon: ```Date now = new Date()```
might be better name for it.

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
```diff
@@ -77,4 +80,15 @@ private static void makeAccessible(final Field field) {
 		}
 	}
 
+	public static List<Field> getDeclaredFieldsUpToParent(Class<?> currentClass, Class<?> exclusiveParent) {
```
junoyoon: What is ```exclusiveParent``` ?
Is it really necessary.

And just ```getDeclaredFieldsIncludingParent``` might be  better name.

* currentClass => current
* parentClass => parent

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
```diff
@@ -77,4 +80,15 @@ private static void makeAccessible(final Field field) {
...
+		List<Field> currentClassFields = newArrayList(currentClass.getDeclaredFields());
+		Class<?> parentClass = currentClass.getSuperclass();
+
+		if (parentClass != null && !(parentClass.equals(exclusiveParent))) {
+			List<Field> parentClassFields = getDeclaredFieldsUpToParent(parentClass, exclusiveParent);
```
junoyoon: Is should be recursive call?

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
```diff
@@ -77,4 +80,15 @@ private static void makeAccessible(final Field field) {
 		}
 	}
 
+	public static List<Field> getDeclaredFieldsUpToParent(Class<?> currentClass, Class<?> exclusiveParent) {
+		List<Field> currentClassFields = newArrayList(currentClass.getDeclaredFields());
```
junoyoon: I think fields is better name than currentClassFields

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
donggyu04: It is optional.
ngirnder may be not use this.

is better remove this?

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
donggyu04: If exist function that get all parent class, i can use loop instead of recursive.
i'll check this.

junoyoon: In java, a class allow only one parent class so we don't need recursive call.. loop is enough.

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
```diff
@@ -33,6 +38,7 @@
 public abstract class ReflectionUtils {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(ReflectionUtils.class);
+	private static Map<Class, List<Field>> fieldCache = new ConcurrentHashMap<>();
```
junoyoon: please put final.

ngrinder-core/src/main/java/org/ngrinder/common/util/ReflectionUtils.java
```diff
@@ -77,4 +83,23 @@ private static void makeAccessible(final Field field) {
 		}
 	}
 
+	public static List<Field> getDeclaredFieldsIncludingParent(Class<?> current) {
+
+		if (fieldCache.containsKey(current)) {
+			return fieldCache.get(current);
+		}
+
+		Class<?> start = current;
```
junoyoon: why is this necessary?
Isn't better to fieldCache.put below here?

junoyoon:
LGTM

PR #374, opened 2019-02-15
Add custom error controller

donggyu04:
Disable default whitelabel error page supported by springboot and add custom error controller.
#### - Before
image

#### - After
image

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/error/controller/DefaultErrorController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/error/controller/DefaultErrorController.java
@@ -1,0 +1,29 @@
+package org.ngrinder.error.controller;
+
+import org.springframework.boot.web.servlet.error.ErrorController;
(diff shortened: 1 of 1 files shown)


ngrinder-controller/src/main/java/org/ngrinder/error/controller/ErrorControllerImp.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import javax.servlet.http.HttpServletRequest;
+
+import static javax.servlet.RequestDispatcher.ERROR_STATUS_CODE;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+
+@Controller
+public class ErrorControllerImp implements ErrorController {
```
junoyoon: Please rename is DefaultErrorController

PR #381, opened 2019-02-28
Apply lastet update merge policy to hazelcast map configuration

donggyu04:
Sometimes when running controller with cluster mode, hazelcast split brain syndrome was occurred.
so i apply lastet update merge policy to resolve it.


junoyoon:
LGTM

PR #383, opened 2019-03-13
Make to use spring default json serialization in controller

donggyu04:
Previously, nGrinder serialize object into json using `gson`.
However, spring support 'jackson' serialization by default.
If we use jackson, codes can be simplified a lot.

* [x] Perftest controller (18)
* [x] AgentManager controller (12)
* [x] MonitorManager controller (2)
* [x] LogMonitor controller (2)
* [x] SystemConfig controller (2)
* [x] FIleEntry controller (7)
* [ ] User controller (9)
* [ ] UserSignUp controller (2)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,7 @@ subprojects {
     hazelcast_version = "3.11.1"
     gson_version = "2.8.0"
     mockito_version = "2.15.0"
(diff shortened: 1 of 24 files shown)


donggyu04:
This feature will proceed after converting to vue.js

PR #399, opened 2019-05-31
Cache user with followers for avoid hibernate lazyinitialization exception

donggyu04:
Currently in switch user mode.
hibernate `Lazyinitializationexception` occurs at below line when `userService.getOne()` return cached user object.

https://github.com/naver/ngrinder/blob/develop/ngrinder-controller/src/main/java/org/ngrinder/infra/spring/UserHandlerMethodArgumentResolver.java#L99

so, i fix to use `getOneWithFollowers`.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/user/service/UserService.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/user/service/UserService.java
@@ -15,6 +15,7 @@ package org.ngrinder.user.service;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
+import org.hibernate.Hibernate;
 import org.ngrinder.common.constant.ControllerConstants;
 import org.ngrinder.infra.config.Config;
 import org.ngrinder.model.PerfTest;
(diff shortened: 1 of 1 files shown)


ngrinder-controller/src/main/java/org/ngrinder/user/service/UserService.java
```diff
@@ -108,13 +112,13 @@ public User getOne(String userId) {
 	 * @return user
 	 */
 	@Transactional
+	@Cacheable(value = CACHE_USER_WITH_FOLLOWERS, key = "#userId")
 	public User getOneWithFollowers(String userId) {
-		User one = userRepository.findOneByUserId(userId);
-		one.getFollowers().size();
-		return one;
+		User user = userRepository.findOneByUserId(userId);
+		Hibernate.initialize(user.getFollowers());
```
junoyoon: what about doing this in ```getOne```

songeunwoo: ```Hibernate.initialize(user.getFollowers());``` this code looks the same as findOne.

http://blog.naver.com/PostView.nhn?blogId=kbh3983&logNo=221170946678

maybe getOne could EntityNotFoundException occur?

donggyu04: @junoyoon @songeunwoo 

Making `getOne()` as eager(followers) may affect other usage points, but it seems not problem.
it's better to have fewer changes than before.

PR #401, opened 2019-06-04
Set factual user as created user before saving perftest

donggyu04:
In user switching mode, set factual user as created user before saving perftest

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/common/service/ModelAspect.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/common/service/ModelAspect.java
@@ -68,16 +68,18 @@ public class ModelAspect {
 				BaseModel<?> model = (BaseModel<?>) object;
 				Date lastModifiedDate = new Date();
 				model.setLastModifiedDate(lastModifiedDate);
+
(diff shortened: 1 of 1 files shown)


ngrinder-controller/src/main/java/org/ngrinder/common/service/ModelAspect.java
```diff
@@ -68,16 +68,18 @@ public void beforeSave(JoinPoint joinPoint) {
...
+
 				User currentUser = userContext.getCurrentUser();
-				Long currentUserId = currentUser.getId();
+				long currentUserId = currentUser.getId();
 
 				model.setLastModifiedUser(userRepository.findOne(idEqual(currentUserId))
 					.orElseThrow(() -> new IllegalArgumentException("No user found with id : " + currentUserId)));
 
 				if (!model.exist() || model.getCreatedUser() == null) {
+					long factualUserId = currentUser.getFactualUser().getId();
```
donggyu04: `getFactualUser()` always return not null value.

https://github.com/naver/ngrinder/blob/develop/ngrinder-core/src/main/java/org/ngrinder/model/User.java#L334

PR #409, opened 2019-06-19
Implement perftest duration slider

donggyu04:
Implement perftest duration slider.
this slider bar works in both directions depending on  time-select and weight.

image


songeunwoo:
LGTM!!

PR #411, opened 2019-06-20
Change gradle build-output directory path for matching with webpack output path

donggyu04:
Change gradle build-output directory path for matching with webpack output path.


junoyoon:
LGTM

PR #421, opened 2019-07-01
Fix mysql security vulnerability

imbyungjun:
- Fix mysql security vulnerability
- Add exclusion columns when migrating


donggyu04:

junoyoon:
good!

PR #428, opened 2019-07-08
Allow log file deletion in security mode

imbyungjun:
[Merge into `master`]

Allow log file deletion in ngrinder secure mode.

Currently, `logback` isn't rolling test logs in secure mode.


donggyu04:

PR #442, opened 2019-08-01
Resolve failed test due to fixed IP address of www.sample.com

imbyungjun:


donggyu04:

PR #444, opened 2019-08-07
PF4J 3.x supports JAR plugins

imbyungjun:
- Doesn't need to use `plugin_cache` directory anymore
- Doesn't need to find jar files by using custom extension finder

> Since the pf4j package name has changed, plugins need to be re-built with the appropriate pf4j version


donggyu04:

ngrinder-controller/src/main/java/org/ngrinder/infra/plugin/extension/NGrinderDefaultPluginManager.java
```diff
@@ -1,64 +1,40 @@
...
-	}
+	private ApplicationContext applicationContext;
 
 	@Autowired
-	public void setSpringExtensionFactory(SpringExtensionFactory extensionFactory) {
-		super.extensionFactory = extensionFactory;
+	public NGrinderDefaultPluginManager(Config config) {
+		super(config.isClustered()
+			? config.getExHome().getPluginsDirectory().toPath()
+			: config.getHome().getPluginsDirectory().toPath());
```
imbyungjun: Change to home directory.

PR #448, opened 2019-08-16
Apply URL decoding when getting file name from URLClassLoader

donggyu04:
#446

File couldn't be found if there were certain characters in the path when getting the file name from `URL` object (from a URLClassLoader). so apply URL decoding to file name.

#### completed test case
* [x] Download agent
* [x] Download monitor
* [x] Test running with monitor and plugins (using network overflow plugin)
* [x] Script CRUD

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/engine/agent/LocalScriptTestDriveService.java
+++ b/ngrinder-core/src/main/java/net/grinder/engine/agent/LocalScriptTestDriveService.java
@@ -34,6 +34,7 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Properties;
 
+import static org.ngrinder.common.util.EncodingUtils.decodePathWithUTF8;
 import static org.ngrinder.common.util.NoOp.noOp;
 
 /**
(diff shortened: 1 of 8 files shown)


ngrinder-core/src/main/java/net/grinder/engine/agent/LocalScriptTestDriveService.java
```diff
@@ -237,11 +238,11 @@ private boolean isRunningOnWas() {
 		return ((URLClassLoader) LocalScriptTestDriveService.class.getClassLoader()).getURLs()[0].getProtocol().equals("jar");
 	}
 
-	private String runtimeClassPath() {
+	private String runtimeClassPath() throws UnsupportedEncodingException {
```
junoyoon: Isn't it better to eat this exception silently?

junoyoon: with try / catch internally.

donggyu04: This exception seems will not occur when using `UTF-8`

when using try/catch internally how about like this?
```
try {
	runtimeClassPath.append(URLDecoder.decode(url.getFile(), "UTF-8")).append(File.pathSeparator);
} catch (UnsupportedEncodingException e) {
	throw new NGrinderRuntimeException("error while resolving runtime class path for script validation check", e);
}
```

junoyoon:
LGTM

PR #451, opened 2019-08-23
Serialize JSON NaN as null

imbyungjun:
- Sometimes cpu usage percentage presents as NaN
- There's no NaN in JSON format, represent it as null

### DIFF
--- a/ngrinder-frontend/src/js/components/perftest/mixin/FormatMixin.vue
+++ b/ngrinder-frontend/src/js/components/perftest/mixin/FormatMixin.vue
@@ -6,11 +6,11 @@
         formatNetwork(format, value) {
             value = value || 0;
             if (value < 1024) {
-                return `${value.toFixed(1)}B `;
+                return `${value.toFixed(1)}B`;
             } else if (value < 1048576) { // 1024 * 1024
-                return `${(value / 1024).toFixed(1)}K `;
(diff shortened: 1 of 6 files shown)


donggyu04:

imbyungjun:
https://github.com/naver/ngrinder/blob/d713428c8d084c231512f9d2484714279feff167/ngrinder-controller/src/main/java/org/ngrinder/agent/controller/AgentManagerApiController.java#L153

to serialize system data model, the number module have to be added to both objectMapper bean and JsonUtils

junoyoon:
Very good!

ngrinder-frontend/src/js/components/perftest/mixin/FormatMixin.vue
```diff
@@ -28,7 +28,10 @@
         }
 
         formatPercentage(format, value) {
-            value = value || 0;
+            if (!value) {
+                return 'N/A';
```
junoyoon: What about give space after A like ``` 'N/A '```

imbyungjun: I tried to add a space after it but, this value is wrapped by `<span/>` tag. So additional white space doesn't affect to html.
image

junoyoon:
LGTM

PR #452, opened 2019-08-23
Apply vuetable-2 for datatable

donggyu04:
Apply [vuetable-2](https://www.npmjs.com/package/vuetable-2) for datatable.
it supports data sorting and pagination.


ngrinder-frontend/src/js/components/agent/List.vue
```diff
@@ -36,123 +36,137 @@
...
-        page = {
-            number: 1,
-            totalPages: 1,
-            size: 10,
+        table = {
+            css: {},
+            renderingData: {
+                data: [],
+                pagination: {
+                    per_page: 10,
```
junoyoon: why does it have different snake case?

junoyoon:
LGTM

ngrinder-frontend/src/js/components/agent/List.vue
donggyu04: vuetable-2 uses snake case like this
https://www.vuetable.com/guide/pagination.html#how-the-pagination-component-works

so I use snake case to table related value.
change to using the camel case?

donggyu04: @junoyoon 
Changed from snake case to carmel case for all tables.
https://github.com/naver/ngrinder/pull/452/commits/53ce08a4746be2545ec376c0cf3457f5dcd35fee#diff-31ce3a93059eae7178230badd1502f68R111

PR #454, opened 2019-08-27
Apply lombok

imbyungjun:
- Apply lombok
- Clean up getters and setters

PR #458, opened 2019-09-09
[WIP] Code Refactoring

imbyungjun:
- Code refactoring and cleanup


donggyu04:

imbyungjun:
This commit makes `addNewUserIntoLocal()` method async. It means that server may enter to unpredicted status. For example, Authentication is succeed but, there might be no svn user directory. 

I will find out some other way to avoid `ModelAspect`

imbyungjun:
I think that security context method would be better than unnecessary async thread execution and wait

imbyungjun:
Merge into https://github.com/naver/ngrinder/pull/454

PR #461, opened 2019-10-01
Use COPY instead of ADD in Dockerfile

gy741:
Hello,

The use of COPY is preferred over ADD when additional functionalities of ADD are not being used (like spraying the contents of a tar in a folder).

Ref: https://docs.docker.com/engine/articles/dockerfile_best-practices/#add-or-copy

Thanks.


donggyu04:

junoyoon:
LGTM

junoyoon:
Thx

PR #462, opened 2019-10-07
Fix minor bug and adjust some css

donggyu04:
1. Adjust agent and perftest list table column size.
2. Remove lineSeparator option from codemirror.

> This version has deploymented  in dev-cluster.


junoyoon:
LGTM

PR #470, opened 2019-10-21
Ensure data ready before rendering

imbyungjun:
- Make data ready before rendering script list, script editor and perftest detail
- Fix bug that cannot refresh on quickstart


junoyoon:
LGTM

PR #477, opened 2019-11-06
Refactor AgentManager

donggyu04:
##### Refactor `AgentManager, AgentManagerService`

- ` AgentManager`: handle attached agent at specific node only.
- `AgentService`: handle whole attached agent.


ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+				readyAgentCnt++;
+			}
+		}
+		return readyAgentCnt;
+	}
+
+	@Override
+	public List<AgentInfo> getAllVisible() {
+		return getAllActive();
+	}
```
junoyoon: What about just delete out ```getAllVisible```

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+
+	@Override
+	public List<AgentInfo> getAllAttached() {
+		return agentInfoStore.getAllAgentInfo();
+	}
+
+	@Override
+	public List<AgentInfo> getLocalAgents() {
+		return getAllAttached();
+	}
```
junoyoon: What about just using getAllAttached()

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+
+	@Override
+	public String createKey(AgentInfo agentInfo) {
+		return agentInfo.getAgentKey();
+	}
+
+	@Override
+	public String createKey(AgentControllerIdentityImplementation agentIdentity) {
+		return createAgentKey(agentIdentity.getIp(), agentIdentity.getName());
+	}
```
junoyoon: What about make these key related method private?

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+
+		for (AgentProcessControlImplementation.AgentStatus status : agentMap.values()) {
+			AgentControllerIdentityImplementation agentIdentity = (AgentControllerIdentityImplementation) status.getAgentIdentity();
+			AgentInfo agentInfo = agentInfoStore.getAgentInfo(createKey(agentIdentity));
+			// check new agent
+			if (agentInfo == null) {
+				agentInfo = new AgentInfo();
+			}
+
+			fillUpAgentInfo(agentInfo, status);
```
junoyoon: Do we need periodic fillup? listener isn't enought?

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: It will be extract to abstract class for backward compatibility.

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: It will be extract to abstract class for backward compatibility.

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: This method called by `AgentStatusUpdateListener` only. not periodically called.

junoyoon: okey

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+	public Set<AgentInfo> getAllAttachedFreeApprovedAgentsForUser(User user) {
+		if (user == null) {
+			return emptySet();
+		}
+
+		String region = config.getRegion();
+		String userId = user.getUserId();
+		return getAllFreeApprovedAgents()
+			.stream()
+			.filter(AgentInfo::isApproved)
```
junoyoon: why do we need isApproved here? getAllFreeApprovedAgents() returns approved agents?

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+			.filter(agentInfo -> StringUtils.equals(region, agentManager.extractRegionKey(agentInfo.getRegion())))
```
junoyoon: what about make this part functor?

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: This method is `IAgentManagerService` interface implementation.
so, can't define an access modifier below public.

`public abstract String createKey(AgentInfo agentInfo);`

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -0,0 +1,489 @@
...
+			agentInfoInDB.setIp(ip);
+			agentInfoInDB.setName(name);
+			agentInfoInDB.setApproved(approve);
+			agentManagerRepository.save(agentInfoInDB);
+		}
+		agentManagerRepository.flush();
+
+		AgentInfo agentInfo = agentInfoStore.getAgentInfo(createAgentKey(ip, name));
+		agentInfo.setApproved(approve);
+		agentInfoStore.updateAgentInfo(agentInfo.getAgentKey(), agentInfo);
```
junoyoon: I hope above three lines as a separate method

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: Yes, ㅅhis seems unnecessary.

junoyoon:
LGTM

PR #478, opened 2019-11-07
Change to billboard js chart

imbyungjun:


ngrinder-controller/src/main/java/org/ngrinder/perftest/service/PerfTestService.java
```diff
@@ -1185,29 +1184,29 @@ public int getMonitorGraphInterval(long testId, String targetIP, int imageWidth)
 					skipCount = 0;
 					String[] datalist = StringUtils.split(line, ",");
 					if ("null".equals(datalist[4]) || "undefined".equals(datalist[4])) {
```
donggyu04: use `NULL_STRING` or "null" 
I think it's good to unite in one way.

imbyungjun: I didn't find out there's a code uses the string value "null". I'm gonna change it to use string constants

ngrinder-frontend/src/js/components/perftest/detail/RampUp.vue
```diff
@@ -159,118 +181,150 @@
...
 <style lang="less" scoped>
+    @import '~billboard.js/dist/billboard.min.css';
+    @import '~billboard.js/dist/theme/insight.min.css';
+
     .ramp-up-container {
         width: 460px;
         margin-left: 18px;
 
+        .enable-checkbox {
+            vertical-align: text-bottom;
```
donggyu04: Why don't you use bootstrap predefined class?

like this `<td class="align-text-bottom">text-bottom</td>`

junoyoon:
LGTM

PR #482, opened 2019-11-19
Exclude global attributes settings to api-controller

donggyu04:
Make `GlobalControllerAdvice` work only on controller that return view (app.html)
> exclude api-controller, download controller


ngrinder-controller/src/main/java/org/ngrinder/home/controller/HealthCheckApiController.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+
+/**
+ * Health check api controller.
+ *
+ * @since 3.5.0
+ */
+@Slf4j
+@RestController
+@GlobalControllerModel
```
junoyoon: Why is this necessary?

donggyu04: It my mistake 👍

PR #486, opened 2019-12-03
Make file name encoded with utf-8 on file download

imbyungjun:
- Make utf-8 file name doesn't break on file download
- Support IE11, Safari, Chrome

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
@@ -401,7 +401,6 @@ public class FileEntryApiController {
 	 */
 	@GetMapping("/download/**")
 	public void download(User user, @RemainedPath String path, HttpServletResponse response) {
-		path = decodePathWithUTF8(path);
 		FileEntry fileEntry = fileEntryService.getOne(user, path);
 		if (fileEntry == null) {
 			LOG.error("{} requested to download not existing file entity {}", user.getUserId(), path);
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
@@ -416,21 +416,11 @@ public class FileEntryApiController {
 		response.setHeader("Content-Disposition", contentDisposition.toString());
 		response.addHeader("Content-Length", "" + fileEntry.getFileSize());
 
-		byte[] buffer = new byte[4096];
-		ByteArrayInputStream fis = null;
-		OutputStream toClient = null;
-		try {
-			fis = new ByteArrayInputStream(fileEntry.getContentBytes());
-			toClient = new BufferedOutputStream(response.getOutputStream());
-			int readLength;
-			while (((readLength = fis.read(buffer)) != -1)) {
-				toClient.write(buffer, 0, readLength);
-			}
+		try (ByteArrayInputStream fis = new ByteArrayInputStream(fileEntry.getContentBytes());
+			OutputStream os = new BufferedOutputStream(response.getOutputStream())) {
+			IOUtils.copy(fis, os);
 		} catch (IOException e) {
 			throw processException("error while download file", e);
-		} finally {
-			IOUtils.closeQuietly(fis);
-			IOUtils.closeQuietly(toClient);
 		}
 	}
 }
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
@@ -18,6 +18,7 @@ import static java.util.stream.Collectors.toList;
 import static org.apache.commons.io.FilenameUtils.getPath;
 import static org.apache.commons.lang3.StringUtils.*;
 import static org.ngrinder.common.util.CollectionUtils.buildMap;
+import static org.ngrinder.common.util.EncodingUtils.decodePathWithUTF8;
 import static org.ngrinder.common.util.EncodingUtils.encodePathWithUTF8;
 import static org.ngrinder.common.util.ExceptionUtils.processException;
 import static org.ngrinder.common.util.NoOp.noOp;
@@ -27,6 +28,8 @@ import static org.ngrinder.common.util.Preconditions.checkNotNull;
 
 import com.nhncorp.lucy.security.xss.XssPreventer;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.IOUtils;
+import org.ngrinder.common.exception.NGrinderRuntimeException;
 import org.ngrinder.common.util.PathUtils;
 import org.ngrinder.common.util.UrlUtils;
 import org.ngrinder.infra.spring.RemainedPath;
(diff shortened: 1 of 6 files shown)


imbyungjun:
image

Currently, it has an issue with utf-8 file name on IE11. (This issue also occurred on master branch)

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
```diff
@@ -385,4 +391,47 @@ public String validate(User user, @RequestBody ScriptValidationParams scriptVali
...
+	/**
+	 * Download file entry of given path.
+	 *
+	 * @param user     current user
+	 * @param path     user
+	 * @param response response
+	 */
+	@GetMapping("/download/**")
+	public void download(User user, @RemainedPath String path, HttpServletResponse response) {
+		path = decodePathWithUTF8(path);
```
junoyoon: Isn't it better to move this to RemainedPathArgumentHandler?

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
```diff
@@ -385,4 +391,47 @@ public String validate(User user, @RequestBody ScriptValidationParams scriptVali
...
+			.build();
+		response.setHeader("Content-Disposition", contentDisposition.toString());
+		response.addHeader("Content-Length", "" + fileEntry.getFileSize());
+
+		byte[] buffer = new byte[4096];
+		ByteArrayInputStream fis = null;
+		OutputStream toClient = null;
+		try {
+			fis = new ByteArrayInputStream(fileEntry.getContentBytes());
+			toClient = new BufferedOutputStream(response.getOutputStream());
```
junoyoon: Use 
```
try(ByteArrayInputStream fis = new ByteArrayInputStream(fileEntry.getContentBytes()), OutputStream toClient = new BufferedOutputStream(response.getOutputStream())) {
...
}
```

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
```diff
@@ -385,4 +391,47 @@ public String validate(User user, @RequestBody ScriptValidationParams scriptVali
...
+		response.setContentType("application/octet-stream; charset=UTF-8");
+		ContentDisposition contentDisposition = ContentDisposition.builder("attachment")
+			.filename(fileEntry.getFileName(), StandardCharsets.UTF_8)
+			.build();
+		response.setHeader("Content-Disposition", contentDisposition.toString());
+		response.addHeader("Content-Length", "" + fileEntry.getFileSize());
+
+		byte[] buffer = new byte[4096];
+		ByteArrayInputStream fis = null;
+		OutputStream toClient = null;
```
junoyoon: Variable name is not symmetric. it's better to name it ```os```

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
```diff
@@ -385,4 +391,47 @@ public String validate(User user, @RequestBody ScriptValidationParams scriptVali
...
+		try {
+			fis = new ByteArrayInputStream(fileEntry.getContentBytes());
+			toClient = new BufferedOutputStream(response.getOutputStream());
+			int readLength;
+			while (((readLength = fis.read(buffer)) != -1)) {
+				toClient.write(buffer, 0, readLength);
+			}
```
junoyoon: See 
https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#copy(java.io.InputStream,%20java.io.OutputStream,%20int)

ngrinder-frontend/src/js/components/script/List.vue
```diff
@@ -218,7 +218,14 @@
...
         downloadScript(path) {
-            window.location.href = `${this.contextPath}/script/download/${path}`;
+            const link = document.createElement('a');
+            link.style.display = 'none';
+            link.href = `${this.contextPath}/script/api/download/${path}`;
+            link.setAttribute('download', path);
+
+            document.body.appendChild(link);
+            link.click();
+            document.body.removeChild(link);
```
junoyoon: What is it for?

imbyungjun: Used it as a workaround for the MIME type warning. It's simpler to use `location.href` but, get a console warning like below.
```
Resource interpreted as Document but transferred with MIME type application/octet-stream ...
```
https://stackoverflow.com/questions/6587393/resource-interpreted-as-document-but-transferred-with-mime-type-application-zip/17452321#17452321

junoyoon: okay

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
imbyungjun: I considered to decode it in the RemainedPathArgumentHandler but, I thought it was danger to decode all the remained path variable.

As I reconsider now, there seems to be no danger. Except the case that decodes remained path variable twice. I'm gonna move this and check the case.

junoyoon: go for it

junoyoon: You can make annotation have decode param. It will make it safe.
```
@RemainedPath(decodeWithUTF8=true)
```

imbyungjun: I inspected side effects when returning UTF-8 decoded RemainedPath variable. 

1. Upload File 
Get the file name from MultiPartFile object, so it doesn't need to be decoded. Get the current working directory as RemainedPath variable but, all directories are english. Also doesn't need to be decoded. 

2. Delete File 
Both v3.5.0 and before than v3.5.0, get file names to delete from request body(POST method). It doesn't need to be decoded. 

3. Download File 
Get target file from given RemainedPath variable. In this case, RemainedPath variable must be decoded. 

As a result, only when download a file must be decode RemainedPath variable. And there seems no side effect when returning UTF-8 decoded RemainedPath variable. 

I will move decoding step to RemainedPathResolver.

junoyoon: By the way, use DELETE method for deleting file.

imbyungjun: Currently, it provides delete API in two methods. One for deleting a file, and the other for deleting multiple files.
https://github.com/naver/ngrinder/blob/48c08dacdc1d6d44c5710841c09787c3aa2e560e/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java#L256
https://github.com/naver/ngrinder/blob/48c08dacdc1d6d44c5710841c09787c3aa2e560e/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java#L370

When changed post to delete method, I think that `@DeleteMapping("/delete")` is redundant. 

How about deleting the first POST method API that used in front-end?

And make the `@DeleteMapping("/**")` API to work with both single file name and file name list. And change to use the delete method in front-end.

imbyungjun: https://github.com/naver/ngrinder/blob/48c08dacdc1d6d44c5710841c09787c3aa2e560e/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java#L370
I didn't recognize this API works for the given RemainedPath, I'm gonna change just POST API to use DELETE method

songeunwoo:
Thanks for your efforts.

PR #490, opened 2020-01-06
Fix log file name for agent command line

imbyungjun:
Currently, command line options for perftest process in the agent is logged only in ${agentName}-${agentNumber}.log file.  
Fix this to ${agentName}-0.log to be sent to controller.


songeunwoo:
LGTM~

donggyu04:
The best solution seems use worker process number to make log file name.
but, this solution cover almost user and looks very simple.

PR #492, opened 2020-01-16
Fix agent status reporting message updation logic

donggyu04:
Currently agent status messages(cpu, mem, rx, etc..) are not being updated properly on running phase.

So, I fix that.


ngrinder-core/src/main/java/net/grinder/console/communication/AgentProcessControlImplementation.java
```diff
@@ -164,11 +149,7 @@ private void updateAgentStatusMap(AgentControllerProcessReportMessage message) {
 	 * @return {@link AgentStatus}
 	 */
 	private AgentStatus getAgentStatus(AgentIdentity agentIdentity) {
-		final AgentStatus existing = m_agentMap.get(agentIdentity);
-		if (existing != null) {
-			return existing;
-		}
-		return new AgentStatus(agentIdentity);
+		return m_agentMap.getOrDefault(agentIdentity, new AgentStatus(agentIdentity));
```
songeunwoo: good~

PR #493, opened 2020-01-16
Use popover to display current agent state

donggyu04:
Use popover to display current agent state.

image


songeunwoo:
LGTM

donggyu04:
Print agent state on popover content and adjust width to content.

image

PR #497, opened 2020-01-20
Feature/add ssl jvm argument

donggyu04:
Add JVM arguments related to SSL before running worker process.

image


donggyu04:
I will make new pr to `master`

PR #498, opened 2020-01-20
Add SSL related JVM arguments before running worker process

donggyu04:
Add SSL related JVM arguments before running worker process.

image


junoyoon:
LGTM

PR #500, opened 2020-01-29
Fix release note

imbyungjun:
* update release note

### DIFF
--- a/RELEASE-NOTE.md
+++ b/RELEASE-NOTE.md
@@ -4,6 +4,7 @@
  is the bug fix release.
 - Changes
   * Make default groovy maven project pom.xml work correctly.
+  * Drop SSLv2Hello protocol.
 
 - Bug Fix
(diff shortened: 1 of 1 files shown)


RELEASE-NOTE.md
```diff
@@ -4,6 +4,7 @@
  is the bug fix release.
 - Changes
   * Make default groovy maven project pom.xml work correctly.
+  * Add SSL JVM arguments.
```
donggyu04: We actually use this arguments when make new worker process.

https://github.com/naver/ngrinder/blob/develop/ngrinder-core/src/main/java/net/grinder/engine/agent/AgentImplementationEx.java#L235

what about `Add SSL related JVM arguments to worker process command`?

imbyungjun: Change to `Drop SSLv2Hello protocol.`

PR #510, opened 2020-03-02
fix Index-Developer Resources-More... Open link error

lingsoul:
fix Index-Developer Resources-More... Open link error


donggyu04:
@lingsoul 
It seems your base branch is not develop.
so many changes in this PR.

PR #515, opened 2020-03-02
fix Index-Developer Resources-More... Open link error

lingsoul:
fix Index-Developer Resources-More... Open link error

PR #516, opened 2020-03-02
Add resource url link to home panel

donggyu04:
Add resource url link to home panel.


donggyu04:
This should merge first before take #515.

imbyungjun:
How about changing here to use prop`seeMoreResourcesUrl`?
https://github.com/naver/ngrinder/blob/171975bf0e0574cb26b9aa7139256f40e0a8ac32/ngrinder-frontend/src/js/components/Home.vue#L21

PR #518, opened 2020-03-03
fix Index-Developer Resources-More... Open link error use new PR

lingsoul:


donggyu04:
@lingsoul 

Thanks :)

PR #519, opened 2020-03-04
Enhance test re-execution speed

donggyu04:
Enhance test re-execution speed by caching distribution files in agent side.


ngrinder-core/src/main/java/org/ngrinder/common/util/StreamUtils.java
```diff
@@ -0,0 +1,20 @@
+package org.ngrinder.common.util;
+
+import java.util.function.Function;
+
+public class StreamUtils {
+	public static <T, R> Function<T, R> checkedExceptionToRuntimeWrapper(ExceptionFunction<T, R> fe) {
```
songeunwoo: The wrapping method is nice to see.
I would like to simply change the method name :)

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
```diff
@@ -86,6 +88,9 @@
 
 	private final Condition eventSyncCondition = new Condition();
 	private ProcessReports[] processReports;
+
+	// It contains cached distribution files md5 checksum from each agents.
+	private List<Set<String>> agentCachedDistFilesMd5List = synchronizedList(new ArrayList<>());
```
songeunwoo: How about use Concurrent Collections rather than Synchronized Collections?
https://javarevisited.blogspot.com/2016/05/what-is-difference-between-synchronized.html

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
```diff
@@ -349,6 +359,26 @@ public String getConsoleIP() {
...
+	}
+
+	/**
+	 * Send Md5 of unnecessary files to agents for refresh agent's distribution cache directory.
+	 *
+	 * @param distFilesMd5 Required file's md5 for currently running test
+	 */
+	public void sendDistFilesMd5ToAgents(Set<String> distFilesMd5) {
+		stream(processReports)
+			.parallel()
```
songeunwoo: I don't want to use parallelStream.
https://multifrontgarden.tistory.com/254

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
```diff
@@ -86,6 +92,9 @@
 
 	private final Condition eventSyncCondition = new Condition();
 	private ProcessReports[] processReports;
+
+	// It contains cached distribution files md5 checksum from each agents.
+	private List<Set<String>> agentCachedDistFilesMd5List = synchronizedList(new ArrayList<>());
```
songeunwoo: How about use Concurrent Collections rather than Synchronized Collections?
https://javarevisited.blogspot.com/2016/05/what-is-difference-between-synchronized.html

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -321,17 +325,47 @@ public synchronized void runAgent(User user, final SingleConsole singleConsole,
...
-	 * @return selected agent.
+	 * @param allFreeAgents available agents
+	 * @param agentCount    number of agents
+	 * @return selected agents.
 	 */
-	private Set<AgentInfo> selectAgent(User user, Set<AgentInfo> allFreeAgents, int agentCount) {
-		Stream<AgentInfo> ownedFreeAgentStream = allFreeAgents.stream().filter(agentInfo -> isOwnedAgent(agentInfo, user.getUserId()));
-		Stream<AgentInfo> freeAgentStream = allFreeAgents.stream().filter(this::isCommonAgent);
+	Set<AgentInfo> selectAgent(User user, Set<AgentInfo> allFreeAgents, int agentCount) {
+		Set<AgentInfo> recentlyUsedAgents = hazelcastService.getOrDefault(CACHE_RECENTLY_USED_AGENTS, user.getUserId(), emptySet());
```
imbyungjun: How about using stream style like before?
Get `recentlyUsedAgents` as a stream and then, concat these all three streams and limit by `agentCount`.

ngrinder-core/src/main/java/org/ngrinder/common/util/StreamUtils.java
donggyu04: Make simple `exceptionWrapper`

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
donggyu04: Using parallel streams elsewhere without a separate thread pool might be a problem.
so modified to work in separate thread pool.

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
donggyu04: When to use SynchronizedList?
1. Since in CopyOnWriteArrayList for every update/modify operations, a new separate cloned copy is created and there is overhead on JVM to allocate memory & merging cloned copy with original copy. Thus, in this case SynchronizedList is better option.
2. When size of Arraylist is large.

When to use CopyOnWriteArrayList?

1. The CopyOnWriteArrayList provides reading without a lock, which means a much better performance if there are more reader threads and write is happening quite low.
2. When the size of Arraylist is small.

In our case it is better to use CopyOnWriteArrayList because the list size is small and only use read operation after the first add to the list.

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
donggyu04: Yes, It better.
It is modified.

ngrinder-core/src/main/java/net/grinder/SingleConsole.java
songeunwoo: Using ForkJoinPool is good :)
If the two performance differences aren't big, I don't want them in parallel.

Can you tell me the difference on average based on five agents?

donggyu04: We are already sending agent start messages in parallel.
So, I used parallel in the same context.

There seems to be no big difference in performance.

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -321,15 +325,43 @@ public synchronized void runAgent(User user, final SingleConsole singleConsole,
...
+	 * @return selected agents.
 	 */
-	private Set<AgentInfo> selectAgent(User user, Set<AgentInfo> allFreeAgents, int agentCount) {
-		Stream<AgentInfo> ownedFreeAgentStream = allFreeAgents.stream().filter(agentInfo -> isOwnedAgent(agentInfo, user.getUserId()));
-		Stream<AgentInfo> freeAgentStream = allFreeAgents.stream().filter(this::isCommonAgent);
-		return concat(ownedFreeAgentStream, freeAgentStream).limit(agentCount).collect(toSet());
+	Set<AgentInfo> selectAgent(User user, Set<AgentInfo> allFreeAgents, int agentCount) {
+		Set<AgentInfo> recentlyUsedAgents = hazelcastService.getOrDefault(CACHE_RECENTLY_USED_AGENTS, user.getUserId(), emptySet());
+
+		Comparator<AgentInfo> agentInfoComparator = (agent1, agent2) -> {
```
imbyungjun, line 342: Naming this comparator as `recentlyUsedPriorityComparator` would be great!

ngrinder-controller/build.gradle
```diff
@@ -108,6 +108,12 @@ task convert_cr_lf {
     }
 }
 
+processResources {
+    filesMatching("ngrinder-sh/agent/run_agent_internal.*") {
+        expand(["ngrinder_core": String.format("lib/ngrinder-core-%s.jar", project.version)])
+    }
+}
```
donggyu04, line 115: Some grinder's class is customized for ngrinder.
so, agent must load ngrinde-core before grinder's

ngrinder-controller/src/main/java/org/ngrinder/agent/service/AgentService.java
```diff
@@ -313,6 +319,16 @@ public synchronized void runAgent(User user, final SingleConsole singleConsole,
...
 		final Set<AgentInfo> necessaryAgents = selectAgent(user, allFreeAgents, agentCount);
 
+		if (hasOldVersionAgent(necessaryAgents)) {
+			for (AgentInfo agentInfo : necessaryAgents) {
+				if (!agentInfo.getVersion().equals(nGrinderVersion)) {
+					update(agentInfo.getIp(), agentInfo.getName());
+				}
+			}
+			throw new NGrinderRuntimeException("Old version agent is detected so, update message has been sent automatically." +
+				"\n Please restart perftest after few minutes.");
```
donggyu04, line 329: image

songeunwoo:
good job~^^

PR #532, opened 2020-03-26
Fix prepare scripts in perftest detail

donggyu04:
Fix case where admin couldn't look user's scripts up in perftest detail.
 - Appended `ownerId` to query string of getting scripts API when admin look normal user's test up.


songeunwoo:
Good job, I'll merge^^

PR #539, opened 2020-04-14
Apply profile to webpack build

donggyu04:
Apply profile to webpack build.

below command  make run webpack under `production` mode

`gradlew clean build -Pprofile=p`
`gradlew clean build -Pprofile=prod`
`gradlew clean build -Pprofile=production`

else webpack run under `development` mode


songeunwoo:
good job^^

PR #540, opened 2020-04-21
Make editor CSS and split panes compatible in Safari

imbyungjun:
* Fix CSS to make editor properly works in Safari browser


songeunwoo:
Good job~

PR #544, opened 2020-04-24
Remove unnecessary vue hook

donggyu04:
If old chart data exist in `{ngrinder-home}/perftest/{user}/{test-id}/report`
Intermittently,  `Report.vue` is rendered old chart.

because, it read old data in created hook before they are deleted.
so, remove problematic hook, chart data is [already handled by bootstrap tab event](https://github.com/naver/ngrinder/blob/develop/ngrinder-frontend/src/js/components/perftest/detail/Detail.vue#L307)


songeunwoo:
Good job, I'll merge :)

PR #545, opened 2020-05-07
Beautify css

donggyu04:
Adjust some css

1. Show popover scroll only when needed.
image

2. Prevent `SmallChart.vue`'s table style affect other components table style.

image


imbyungjun:
During development, I often saw that crashed table style, the second issue.
But, it was quite hard to debug and to reproduce. 

Such a great job. you've solved it!

songeunwoo:
goob job~

PR #557, opened 2020-05-21
Implement bi-directional agent

imbyungjun:
- Implement bi-directional agent
- Controller has to initiate agent connection


songeunwoo:
Good work :) 👍

PR #571, opened 2020-06-12
Prevent zombie agents from remaining in distributed cache

imbyungjun:
- fix hazelcast map config
- fix timer thread not to be shutdown


songeunwoo:
very good~!!

PR #574, opened 2020-06-13
Add today icon on the perftest created today

imbyungjun:
- simplify to highlight today only, not yesterday or earlier


songeunwoo:
good job :)

PR #576, opened 2020-06-13
Make detail report more natural

donggyu04:
Make detail report more natural.

### DIFF
--- a/ngrinder-frontend/src/js/components/perftest/report/menu/PerfTest.vue
+++ b/ngrinder-frontend/src/js/components/perftest/report/menu/PerfTest.vue
@@ -8,45 +8,47 @@
             </button>
         </div>
 
-        <div v-show="tpsChart">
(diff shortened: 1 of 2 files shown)


ngrinder-frontend/src/js/components/perftest/report/menu/PerfTest.vue
```diff
@@ -130,7 +145,8 @@
             &.bigchart, &.chart {
                 border: 1px solid #c4c4c4;
                 height: 300px;
-                min-width: 615px;
+                min-width: 960px;
+                max-width: 960px;
```
songeunwoo, line 149: I think it would be as simple as 960 in width.
Why did you declare min-width and max-width separately?

donggyu04: Fix that.

PR #584, opened 2020-06-19
Fix agent list pagination

imbyungjun:
- reset page when region filter is changed


songeunwoo:
It's a simple fix, so I just merge.

PR #585, opened 2020-06-22
Fix bugs of github storage

donggyu04:
Fix some bugs.

* [x] 1. Groovy maven project isn't properly checked.


ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
```diff
@@ -37,6 +37,7 @@
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static java.util.concurrent.TimeUnit.MINUTES;
```
imbyungjun, line 40: please check this static import is needed

songeunwoo:
good work :)

PR #588, opened 2020-06-30
Remove XSS filter

imbyungjun:
- remove XSS filter
- prevent XSS by using vue.js `v-text`
- use `v-html` only in system-generated contents not user-generated contents
- let admin to use html freely on the announcement

There are four places that user generated contents can be rendered.
1. **Perftest list**
perftest name popover contains perftest description and comment.
=> change popover not to use data-html.
2. **Perftest detail report**
detail report page renders perftest description and comment.
=> change to use `v-text`
3. **User list**
user list table renders user description.
=> already use `v-text`
4. **Script list**
script list table renders file description.
=> already use `v-text`


songeunwoo:
good job :)

PR #593, opened 2020-07-06
Add editor link to validation result

donggyu04:
1. Add editor link to validation result. when clicked that, focus moves to the corresponding line in the editor.

image
image


songeunwoo:
good~!!

PR #594, opened 2020-07-06
Remove agent configuration overwrite option

donggyu04:
1. If `__agent.conf` exists, it is always overwritten & delete.
2. Remove `agent -o option`.


songeunwoo:
Please change the agent description later.

https://github.com/naver/ngrinder/wiki/Agent-Configuration-Guide

PR #601, opened 2020-07-13
Remove vue-router-referer

donggyu04:
Remove`vue-router-referer` module.


junoyoon:
Pr builder is broken. Is it relevant to this pr?

PR #607, opened 2020-07-27
Prevent overflow of user information modal

donggyu04:
Prevent overflow of user information modal.

1. Before

image

2. After

image


songeunwoo:
LGTM

PR #608, opened 2020-07-27
Change digest of cached files

donggyu04:
Currently, when determine if the file is cached or not, ngrinder only use md5 checksum.

Therefore, if you run test using same script after changing the path, ngrinder controller recognizes that as a cached file and doesn't send it to the agent.

To resolve that, I changed file's digest format to `{relative path from base directory}:{md5 checksum}`.

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/engine/agent/FileStore.java
+++ b/ngrinder-core/src/main/java/net/grinder/engine/agent/FileStore.java
@@ -31,6 +31,7 @@ import net.grinder.util.Directory;
 import net.grinder.util.FileContents;
 import net.grinder.util.StreamCopier;
 import org.apache.commons.io.FileUtils;
+import org.ngrinder.common.exception.NGrinderRuntimeException;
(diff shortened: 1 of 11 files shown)


ngrinder-core/src/main/java/net/grinder/engine/agent/FileStore.java
```diff
@@ -144,7 +144,13 @@ public void handle(RefreshCacheMessage message) {
...
 						cachedFiles
 							.stream()
-							.filter(file -> !isRequiredFile(requiredFilesDigest, file))
+							.filter(file -> {
+								try {
+									return !requiredFilesDigest.contains(getFileDigest(cacheDir, file));
+								} catch (IOException e) {
+									throw new NGrinderRuntimeException(e);
+								}
+							})
```
imbyungjun, line 153: It would be better to change `isRequiredFile()` method to simplify lambda expr. Or remove the `isRequireFile()` method that is unused.

PR #637, opened 2020-08-19
Remove unnecessary personal information

donggyu04:
1. Remove mobile phone column.
2. Add log to user API (save, update, delete)


ngrinder-controller/src/main/java/org/ngrinder/user/controller/UserApiController.java
```diff
@@ -173,15 +176,13 @@ public void save(User user, @RequestBody User updatedUser) {
...
+		save(user.getUserId(), updatedUser);
 	}
 
-	private User save(User user) {
-		if (StringUtils.isBlank(user.getPassword())) {
-			return userService.saveWithoutPasswordEncoding(user);
-		} else {
-			return userService.save(user);
-		}
+	private User save(String modifier, User user) {
```
junoyoon, line 182: I don't understand.. why the first parameter name is modifier?

donggyu04: I tried to indicate who had modified(save, update) the user information.

songeunwoo (approved):

PR #647, opened 2020-09-02
change param length from 50 to 256

colin-lee:
image


donggyu04:
@colin-lee 

We had been fixing this feature and your changes looks not enough to meet our needs.
see #652.

Thanks for your contribution and sorry about closing.

Issue #565, opened 2020-06-09
remove parameter validation for advanced test configuration


MahatmaFatalError:
The limitation to 50 chars and only limited special character is too limited. It prevents use cases like having **json** or **xml** as a parameter.
<img width="709" alt="Bildschirmfoto 2020-06-09 um 12 33 51" src="https://user-images.githubusercontent.com/776825/84137888-a3785f00-aa4d-11ea-95ca-7257b3cc1563.png">

donggyu04:
@MahatmaFatalError 

Thanks for your comments.
I'll consider applying it to the next version.

> PR is always open :)

PR #651, opened 2020-09-08
Remove too strict parameter validation to allow json and xml params

MahatmaFatalError:
See Issue https://github.com/naver/ngrinder/issues/565
Encreased length is handled here https://github.com/naver/ngrinder/pull/647

Caution: I don't have vue skills. This is a test to check if the regex is working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ package-lock.json
 /atlassian-ide-plugin.xml
 src/main/generated/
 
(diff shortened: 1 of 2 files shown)


donggyu04:
@MahatmaFatalError 

We had been fixing this feature and your changes looks not enough to meet our needs.
see #652.

Thanks for your contribution and sorry about closing.

PR #652, opened 2020-09-11
Support Json for perftest param configuration

donggyu04:
1. Support JSON format for perftest's param configuration.
    - When `input param` element has focus, It opens modal for modification.
    - It uses beautify format in modification and uses uglify format in command line.
    - I do not use system environment to send param to worker process. because It is immutable in runtime and JAVA command line(using -D) can be long enough to send param.


ngrinder-frontend/src/js/main.js
```diff
@@ -66,6 +67,8 @@ axiosInstance.interceptors.request.use(config => {
     return config;
 });
 
+window.jsonlint = jsonlint;
```
imbyungjun, line 70: I think that adding a comment that `window.jsonlint` is used by codemirror json lint addon would make easier to understand why does it need to be declared.

junoyoon:
It's very old issue not discussed. I'm closing.

PR #658, opened 2020-09-24
UI improvement

donggyu04:
1. Add button for refreshing agent logs
   - To avoid expensive refresh.

2. Resize column of agent list table
   - Column size of agent list table is fixed.


ngrinder-controller/src/main/resources/messages_kr.properties
```diff
@@ -235,6 +235,7 @@ perfTest.report.downloadCSV=CSV \uB2E4\uC6B4\uB85C\uB4DC
 perfTest.report.message.leaveComment=\uD14C\uC2A4\uD2B8 \uCF54\uBA58\uD2B8\uAC00 \uB4F1\uB85D\uB418\uC5C8\uC2B5\uB2C8\uB2E4.
 perfTest.report.message.leaveComment.error=\uCF54\uBA58\uD2B8 \uB4F1\uB85D\uC5D0 \uC2E4\uD328\uD588\uC2B5\uB2C8\uB2E4.
 perfTest.report.message.fetch.basicReport.error=\uB9AC\uD3EC\uD305 \uC804\uBCF4\uB97C \uAC00\uC838\uC624\uB294\uB370 \uC2E4\uD328\uD588\uC2B5\uB2C8\uB2E4.
+perfTest.report.message.fetch.agentLogs.error=에이전트 로그를 가져오지 못했습니다
```
songeunwoo, line 238: why korean?

donggyu04: We are using `UTF-8` as a project default encoding.
so it is not problem.

korean is more readable than unicode, I plan to change all unicode to korean later.

songeunwoo: ok

songeunwoo (approved):

PR #667, opened 2020-10-08
Fix case unintended reset of followers at login

donggyu04:
Fix case unintended reset of followers at login


junoyoon:
Could u explain the problems?

donggyu04:
@junoyoon 

We added below line at implementaton of LDAP plugin.

https://github.com/naver/ngrinder/blob/develop/ngrinder-controller/src/main/java/org/ngrinder/security/NGrinderAuthenticationProvider.java#L127

So, user information is updated each login time for synchronize login resource(LDAP) and our DB.
then, loaded user(created at login) doesn't have followers information therefore it resets the existing followers data.

junoyoon (approved):

PR #668, opened 2020-10-08
Move cache of recently used agents to distribution map

donggyu04:
Move cache of recently used agents to distribution map.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
@@ -149,6 +149,7 @@ public class DynamicCacheConfig implements ClusterConstants {
 		cm.addDistMap(DIST_MAP_NAME_SAMPLING, 15);
 		cm.addDistMap(DIST_MAP_NAME_MONITORING, 15);
 		cm.addDistMap(DIST_MAP_NAME_AGENT, 10);
+		cm.addDistMap(DIST_MAP_CACHE_RECENTLY_USED_AGENTS, 1 * DAY);
 
 		cm.addDistCache(CACHE_USERS, 30, 300);
 		cm.addDistCache(CACHE_FILE_ENTRIES, 1 * HOUR + 40 * MIN, 300);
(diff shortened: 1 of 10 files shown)


songeunwoo (approved):

ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
```diff
@@ -149,6 +149,7 @@ public CacheConfigHolder cacheConfigMap() {
 		cm.addDistMap(DIST_MAP_NAME_SAMPLING, 15);
 		cm.addDistMap(DIST_MAP_NAME_MONITORING, 15);
 		cm.addDistMap(DIST_MAP_NAME_AGENT, 10);
+		cm.addDistMap(CACHE_RECENTLY_USED_AGENTS, 1 * DAY);
```
junoyoon, line 152: Please rename CACHE_RECENTLY_USED_AGENTS to DIST_ MAP_RECENTLY_USED_AGENTS

junoyoon:
Please describe the problem when you fix the bugs.

donggyu04:
@junoyoon 

We are using hazelcast dist map to store the data of recently used agents.
but in configuration, it's set to a local cache.

Simply the expire time that I set is not being applied. (T_T)

ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
donggyu04: Fixed

ngrinder-controller/src/main/java/org/ngrinder/infra/config/DynamicCacheConfig.java
```diff
@@ -149,6 +149,7 @@ public CacheConfigHolder cacheConfigMap() {
 		cm.addDistMap(DIST_MAP_NAME_SAMPLING, 15);
 		cm.addDistMap(DIST_MAP_NAME_MONITORING, 15);
 		cm.addDistMap(DIST_MAP_NAME_AGENT, 10);
+		cm.addDistMap(DIST_MAP_CACHE_RECENTLY_USED_AGENTS, 1 * DAY);
```
junoyoon, line 152: Could u rename all const here? 
I think it's not consistent b/w local and remote cache...

For example, I think the following const is better.

* DIST_CACHE_SAMPLING
* DIST_CACHE_RECENTLY_USED_AGENTS
* LOCAL_CACHE_USERS

Please fix them by seperate commit.

songeunwoo (approved):

PR #677, opened 2020-10-21
[WIP] Support HTTP2

imbyungjun:
- Support HTTP2 by using [Java HTTP Client](https://openjdk.java.net/groups/net/httpclient/intro.html) (From Java 11) or [Apache HTTP Client](https://hc.apache.org/httpcomponents-client-5.0.x/quickstart.html) (From httpcomponents 5.0.x)


ngrinder-groovy/src/main/java/org/ngrinder/http/RealHTTPRequest.java
```diff
@@ -0,0 +1,217 @@
...
+
+import javax.net.ssl.HostnameVerifier;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.security.Security;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+public class RealHTTPRequest {
```
junoyoon, line 43: What about InternalHttpRequest?

imbyungjun: We have to hide this class internally to avoid byte code instrument. So, I think that `InternalHttpRequest` makes sense. I'll change it.

ngrinder-groovy/src/main/java/net/grinder/scriptengine/groovy/junit/GrinderRunner.java
```diff
@@ -300,8 +295,8 @@ protected Statement withBeforeThread(Statement statement) {
 	protected void registerRunNotifierListener(RunNotifier notifier) {
 		notifier.addFirstListener(new RunListener() {
 			@Override
-			public void testStarted(Description description) {
-
+			public void testFinished(Description description) throws Exception {
```
donggyu04, line 298: @imbyungjun 

This hook is `Called when an atomic test has finished`
If test script has `@Test` function more than 2, Isn't the connection closed in every function call? and user uses a `Jython` script. This hook will not be called.

I think connection reset logic should be in `ThreadLifeCycleListener.beginRun()` or `ThreadLifeCycleListener.endRun()`

donggyu04: @imbyungjun 

Check `GrinderThread.run()`, The hook is called before and after `workerRunnable.run()`

```
     int currentRun;

...

      for (currentRun = 0;
           numberOfRuns == 0 || currentRun < numberOfRuns;
           currentRun++) {

        m_context.setCurrentRunNumber(currentRun);

        m_context.fireBeginRunEvent();

        try {
          workerRunnable.run();
        }
        catch (ScriptExecutionException e) {
          final Throwable cause = e.getCause();

          if (cause instanceof ShutdownException ||
              cause instanceof Sleeper.ShutdownException) {
            m_logger.info(m_context.getLogMarker(), "shut down");
            break;
          }

          m_logger.error(m_context.getLogMarker(),
                         "Aborted run: " + e.getShortMessage(),
                         e);
        }

        m_context.fireEndRunEvent();
      }

...

      m_context.setCurrentRunNumber(-1);
```

donggyu04: @imbyungjun 

Old `HTTPRequest` seems to reset the connection in [here](https://github.com/naver/ngrinder/blob/feature/support-http2/ngrinder-runtime/src/main/java/net/grinder/plugin/http/HTTPPluginThreadState.java#L108)

imbyungjun: @donggyu04 
Thank you for letting me know it. Reset on `beginRun`(as it was) would be great.

ngrinder-runtime/src/main/java/org/ngrinder/http/HTTPRequest.java
```diff
@@ -93,30 +97,56 @@ private static String getJvmVersion() {
...
 			.collect(joining("&"));
 	}
 
-	public HTTPResponse POST(String url, RequestBody body, Headers headers) {
-		Request request = new Request.Builder()
+	private Request getRequest(String url, String method, RequestBody body, Headers headers) {
+		return new Request.Builder()
 			.url(url)
-			.post(body)
+			.method(method, body)
```
imbyungjun, line 147: Such a good way 👍

PR #695, opened 2020-12-03
Add transparent ball anime gif

junoyoon:
I made the ball anime transparent so that I looks good in table


donggyu04:
@junoyoon 

Thank you but, the border of the ball looks a little weird.

ezgif-6-b3ebbe9009f0

junoyoon:
Insn't it better than before? :-)

donggyu04:
@junoyoon 

Yes, It's better :)

donggyu04 (approved):

junoyoon:
The reason why the border is weird is... that it doesn't have any background and shadow.. so that the border can not be smoother than this. (anti-aliasing..)

PR #707, opened 2020-12-18
Increase agent die-hard schedule interval

donggyu04:
Increase agent die-hard schedule interval from 2ms to 2s.


imbyungjun (approved):

ngrinder-controller/src/main/java/org/ngrinder/perftest/service/samplinglistener/AgentDieHardListener.java
```diff
@@ -9,7 +9,7 @@
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
- * limitations under the License. 
+ * limitations under the License.
```
junoyoon, line 12: I think we should resolve this all kind of fix in next pr.

junoyoon (approved):

PR #714, opened 2020-12-22
Fix validation error when using AppClassLoader in over JDK9

donggyu04:
From JDK9 the application class loader is no longer an instance of `java.net.URLClassLoader`
Therefore in a certain environment(run on Intellij), It emits `ClassCastException` when validating script.

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/util/AbstractGrinderClassPathProcessor.java
+++ b/ngrinder-core/src/main/java/net/grinder/util/AbstractGrinderClassPathProcessor.java
@@ -18,10 +18,13 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 
 import java.io.File;
+import java.lang.management.ManagementFactory;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Stream;
 
 import static org.ngrinder.common.util.CollectionUtils.newArrayList;
 import static org.ngrinder.common.util.EncodingUtils.decodePathWithUTF8;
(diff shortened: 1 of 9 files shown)


ngrinder-core/src/main/java/net/grinder/util/AbstractGrinderClassPathProcessor.java
```diff
@@ -224,22 +227,41 @@ public String buildForemostClasspathBasedOnCurrentClassLoader(Logger logger) {
...
+		URL[] urls = urlsFromClassLoader(AbstractGrinderClassPathProcessor.class.getClassLoader());
 		StringBuilder builder = new StringBuilder();
-		for (URL each : urLs) {
+		for (URL each : urls) {
 			builder.append(decodePathWithUTF8(each.getFile())).append(File.pathSeparator);
 		}
 		return filterPatchClassPath(builder.toString(), logger);
 	}
 
+	public static URL[] urlsFromClassLoader(ClassLoader classLoader) {
```
junoyoon, line 238: is it the appropriate method name?

I think it's better to name it like...
getClassPaths?

donggyu04: @junoyoon 

Renamed the method. and I removed all the codes that cast class loader to `URLClassLoader`.

ngrinder-core/src/main/java/org/ngrinder/common/util/TypeConvertUtils.java
```diff
@@ -0,0 +1,29 @@
...
+
+	/**
+	 * Convert the given object to the inferred return type.
+	 *
+	 * @param object object to be converted.
+	 * @param <T>    converted type
+	 * @return converted object.
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T> T cast(Object object) {
```
songeunwoo, line 26: nice

songeunwoo (approved):

imbyungjun (approved):

Issue #657, opened 2020-09-23
why V3.5.1 Script Validation Error？


huangchuanyu:
script/api/validate 500
error message："Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: Could not initialize class org.ngrinder.common.util.SystemInfoUtils"

donggyu04:
@huangchuanyu 

Could you describe your situation in more detail?

1.  OS
2. JDK version
3. How to run ngrinder
4. Your test script (if possible)
5. Steps to reproduce the bug

huangchuanyu:
windows10  
jdk 1.8.0_717
run through java -jar
Just a normal post request，and performance testing is available

18516329677:
this issue still exist in version 3.5.2, i run ngrinder with idea or java commandline, and when i try to validate script in script detail, the validate api will show 500: java.lang.NoSuchMethodError: com.sun.jna.platform.win32.Kernel32.VerSetConditionMask(JIB)J. but when i run ngrinder with tomcat, validate is ok

kdw9502:
this issue still exist in 3.5.3 for all scripts

/script/api/validate 

return this message

{
  "message" : "Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: Could not initialize class org.ngrinder.common.util.SystemInfoUtils",
  "success" : false,
[...]

PR #716, opened 2020-12-22
Fix script validation error in windows

donggyu04:
#657

Removed old version `jna-platform` to avoid loading competition.


junoyoon (approved):

junoyoon:
Ah... is it wip?

donggyu04:
@junoyoon 

It has not problem in common usecase.
After more testing, If there is a side effect, I will make another PR :)

colin-lee:
same question on 3.5.6, docker

PR #733, opened 2021-01-07
Change default H2 database url in easy clustering mode

donggyu04:
If use implicitly relative path to H2 database url, It emits error and required additional options.
To make UX easier,  I change the default H2  database url to use relative path of user home when user running on easy clustering mode.

image


donggyu04:
Move to #739

PR #737, opened 2021-01-08
[WIP] Support HTTP/2.0 request

imbyungjun:
- Support HTTP2 request by using Apache HTTP Client 5

### DIFF
--- a/ngrinder-runtime/src/main/java/org/ngrinder/http/HttpResponse.java
+++ b/ngrinder-runtime/src/main/java/org/ngrinder/http/HttpResponse.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2012-present NAVER Corp.
+ *
+ * This file is part of The nGrinder software distribution. Refer to
+ * the file LICENSE which is part of The nGrinder distribution for
+ * licensing details. The nGrinder distribution is available on the
+ * Internet at https://naver.github.io/ngrinder
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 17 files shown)


ngrinder-runtime/src/main/java/org/ngrinder/http/HttpResponse.java
```diff
@@ -0,0 +1,63 @@
...
+		this.response = response;
+	}
+
+	public byte[] getBodyBytes() {
+		return response.getBodyBytes();
+	}
+
+	public Map<?, ?> getBodyJson() {
+		return JsonUtils.deserialize(getBodyBytes());
+	}
```
junoyoon, line 42: Do we really need body json code in HttpResponse level?
What about provide some mechanism for user to get json in their favorite way?

such as..

```
   response.getBody(asJson());
```
We can provide ```asJson ``` using groovy json library.
If we do this, we can just take out the jackson dependency in runtime.

imbyungjun: Groovy and Jython both supports json parser as built-in library. They provide a convert method `String` to `Map`(=json).

If we provide method gets the converter as a parameter like below, we can support json as a way integrated with each script engine.

**HttpResponse**
```java
public <T> T getBody(Function<String, T> converter) {
	return converter.apply(getBodyText());
}
```

**Groovy**
```groovy
import java.util.function.Function

import groovy.json.JsonSlurper

def jsonSlurper = new JsonSlurper()
def asJson = new Function<String, Map<?, ?>>() {
	public Map<?, ?> apply(String s) {
		return jsonSlurper.parseText(s)
	}
};

...
def json = httpResponse.getBody(asJson())
```

**Jython**
```python
from java.util.function import Function

import json

class JsonConverter (Function):
	def apply(self, s):
		return json.loads(s)

asJson = JsonConverter()

...
json = httpResponse.getBody(asJson())
```

It seems simple and sufficient to meet user requirements. I'm gonna remove Jackson dependency in `ngrinder-runtime`.

donggyu04:
Move to #749

PR #738, opened 2021-01-11
Add AJP connector configuration

donggyu04:
Add AJP connector configuration. You can optionally enable when you need it.


junoyoon:
I don't think it's necessary.
Instead, Isn't it better to migrate current apache to nginx in our env?

donggyu04:
@junoyoon 

Yes, nginx looks better and It will be rarely used by other users too.

PR #740, opened 2021-01-11
Support dynamic modification of script template

donggyu04:
In some cases, users want to modify the script template dynamically. so that I changed the way how ngrinder get the template files from resources.

1. When ngrinder is started, the `script_template` resources are copied to under `NGRINDER_HOME`. if `script_template` folder already exists in `NGRINDER_HOME` this step is skipped.
2. ngrinder get template resources from `NGRINDER_HOME/script_template` not from classpath:

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
@@ -23,6 +23,7 @@ import org.ngrinder.common.constant.ClusterConstants;
 import org.ngrinder.common.constant.ControllerConstants;
 import org.ngrinder.common.constants.InternalConstants;
 import org.ngrinder.common.exception.ConfigurationException;
+import org.ngrinder.common.exception.NGrinderRuntimeException;
 import org.ngrinder.common.model.Home;
 import org.ngrinder.common.util.FileWatchdog;
 import org.ngrinder.common.util.PropertiesKeyMapper;
@@ -51,10 +52,13 @@ import java.util.Properties;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static net.grinder.util.NoOp.noOp;
-import static org.apache.commons.io.FileUtils.readFileToString;
+import static org.apache.commons.io.FileUtils.*;
 import static org.apache.commons.lang.StringUtils.isEmpty;
 import static org.ngrinder.common.constant.DatabaseConstants.PROP_DATABASE_UNIT_TEST;
 import static org.ngrinder.common.constants.GrinderConstants.GRINDER_SECURITY_LEVEL_NORMAL;
+import static org.ngrinder.common.model.Home.PATH_SCRIPT_TEMPLATE_DIRECTORY;
+import static org.ngrinder.common.util.FileUtils.copyResourceToFile;
+import static org.ngrinder.common.util.PathUtils.getSubPath;
 import static org.ngrinder.common.util.Preconditions.checkNotNull;
 
 /**
@@ -124,6 +128,7 @@ public class Config extends AbstractConfig implements ControllerConstants, Clust
 			home.init();
 			exHome = resolveExHome();
 			copyDefaultConfigurationFiles();
+			copyScriptTemplates();
 			loadInternalProperties();
 			loadProperties();
 			initHomeMonitor();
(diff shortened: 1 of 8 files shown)


ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
```diff
@@ -130,6 +143,10 @@ private void createFileEntries(User user, String path, String name, String url,
 					fileEntry.getProperties().put("targetHosts", EMPTY);
 				}
 				getFileEntryRepository().save(user, fileEntry, UTF_8.name());
+			} catch (FileNotFoundException e) {
+				copyResourceToFile(getScriptTemplateDirectoryPath() + "/" + scriptTemplatePath, new File(scriptTemplateDirectoryPath + "/" + scriptTemplatePath));
+				throw processException(format("Script template file '%s' is not exist in '%s'\n" +
+					"The file is automatically created. Try again.", scriptTemplatePath, "{NGRINDER_HOME}" + getScriptTemplateDirectoryPath()), e);
```
songeunwoo, line 149: I think the message is too difficult for users.
Any other easy messages?

junoyoon:
Do we really need error message? What about just copying templates if not exists during ngrinder start up time?

ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
```diff
@@ -21,18 +21,22 @@
 package org.ngrinder.script.handler;
 
 import org.apache.commons.io.IOUtils;
+import org.h2.store.fs.FileUtils;
```
junoyoon, line 24: This is very weird FileUtils.. please use the others.

ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
```diff
@@ -92,10 +98,16 @@ public boolean prepareScriptEnv(User user, String path, String fileName, String
...
-		createFileEntries(user, path, name, url, scriptContent);
-		if (createLib) {
-			createLibraryDirectory(user, path);
+		try {
+			createFileEntries(user, path, name, url, scriptContent);
+			if (createLib) {
+				createLibraryDirectory(user, path);
+			}
+		} catch (NGrinderRuntimeException e) {
+			getFileEntryRepository().delete(user, newArrayList(path));
```
junoyoon, line 109: What about using just new ArrayList()?
I think we should avoid google guava's utils as much as possible if the matching feature exist in next JDK version.

donggyu04: > Do we really need error message? What about just copying templates if not exists during ngrinder start up time?

Yes, I changed to copy resources if they don't exist only at the start up time.

junoyoon (dismissed):

ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
```diff
@@ -262,6 +265,26 @@ protected void copyDefaultConfigurationFiles() throws IOException {
 		home.copyFrom(resources);
 	}
 
+	private void copyScriptTemplates() {
+		File templateResourceDirectory = new File(getClass().getResource("/script_template").getFile());
```
donggyu04, line 269: It is not working when running on executable jar
https://stackoverflow.com/a/25873705/8074178

I should use Inputstream to copy the folder or consider other ways.

ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java
```diff
@@ -262,6 +267,31 @@ protected void copyDefaultConfigurationFiles() throws IOException {
...
+			try {
+				ResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+				Resource[] resources = resolver.getResources("classpath*:script_template/**/*.*");
+				for (Resource resource : resources) {
+					String scriptTemplatePath = home.getScriptTemplateDirectory().getAbsolutePath();
+					scriptTemplatePath += getSubPath(PATH_SCRIPT_TEMPLATE_DIRECTORY, resource.getURL().getPath());
+					copyResourceToFile(resource, new File(scriptTemplatePath));
+				}
+			} catch (IOException e) {
+				throw new NGrinderRuntimeException("Error while copy script templates.", e);
```
junoyoon, line 282: copy => copying

donggyu04: Fixed

https://github.com/naver/ngrinder/blob/a2d2ef5be0a102aab500446eccac41c36904d9c2/ngrinder-controller/src/main/java/org/ngrinder/infra/config/Config.java#L282

junoyoon (approved):

PR #748, opened 2021-01-28
Make grinder utils work on Jython performance test

imbyungjun:
- Cause `ngrinder-groovy` module doesn't included when running jython performance test,
   grinder utils have to be moved to `ngrinder-runtime` module


junoyoon:
good!!

junoyoon (approved):

PR #754, opened 2021-02-01
Add ES5 checking task to build cycle

donggyu04:
1. Check es5 syntax at build time.
2. transpiled vuejs-datepicker because it contains the `const` keyword.

> const is fully supported by IE11 but caught by es-check. There seems to be no option to skip certain keywords.


songeunwoo (approved):

junoyoon:
build this

junoyoon (approved):

PR #759, opened 2021-02-15
Add custom user search filter for LDAP

kwangil-ha:


junoyoon:
please rebase instead of merging from develop

PR #761, opened 2021-02-16
Add custom user search filter for LDAP

kwangil-ha:


imbyungjun:
Thanks for PR and your interest in the default LDAP plugin.
But, are you sure that these changes are adopt to common LDAP use case?

I don't understand  what you wanna do with `{login}`.

PR #769, opened 2021-03-02
Remove unnecessary json parser

donggyu04:
We are using [jsonc](https://www.npmjs.com/package/jsonc) module to parsing json string in FE.
this module recently making the below error and seems to be no longer updated.
so I drop the module and using javascript built-in object.


imbyungjun (approved):
Good!

songeunwoo (approved):

PR #749, opened 2021-01-28
Support HTTP2 request with Apache HttpCore 5.0

imbyungjun:
- Implement HTTP request wrapper with Apache HttpCore 5.0
- Use simple connection pool to fit in nGrinder

PR #775, opened 2021-03-02
Support connection reset configuration

donggyu04:
Support connection reset option as an advanced configuration.
this option is supported by other tools such as JMeter, Gatling.

> :heavy_exclamation_mark: Please consider merging after https://github.com/naver/ngrinder/pull/749


ngrinder-controller/src/main/resources/messages_en.properties
```diff
@@ -207,6 +207,7 @@ perfTest.config.process=Processes
 perfTest.config.thread=Threads
 perfTest.config.ignoreTooManyError=Ignore Errors
 perfTest.config.ignoreTooManyError.help=If you want to continue testing even if too many errors occur. Please check this.
+perfTest.config.connectionReset=Connection reset after one test cycle
```
junoyoon, line 210: Connection reset on each test run.

ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
```diff
@@ -184,6 +185,10 @@ public String buildJVMArgumentWithoutMemory() {
 			addNativeLibraryPath(jvmArguments);
 		}
 
+		if (properties.getBoolean(GRINDER_PROP_CONNECTION_RESET, false)) {
+			jvmArguments.append(" -Dngrinder.connection.reset.after.one.test.cycle=true ");
```
junoyoon, line 189: -Dngrinder.connection.reset.on.each.test.run

ngrinder-runtime/src/main/java/net/grinder/plugin/http/HTTPPluginThreadState.java
```diff
@@ -0,0 +1,155 @@
...
+
+		final HTTPConnectionWrapper existingConnectionWrapper =
+			m_httpConnectionWrappers.get(keyURI);
+
+		if (existingConnectionWrapper != null) {
+			return existingConnectionWrapper;
+		}
+
+		final HTTPPluginConnectionDefaults connectionDefaults =
+			HTTPPluginConnectionDefaults.getConnectionDefaults();
```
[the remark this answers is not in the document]
donggyu04: This class is a copy of grinder's `HTTPPluginThreadState` and other copied files are also using an origin file format so, I  did not modify it.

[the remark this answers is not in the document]
junoyoon: Okey Dokey

ngrinder-controller/src/main/resources/messages_en.properties
donggyu04: The reason I used the word `test cycle` is `each test` might be confusing in meaning like below.

1. Is it reset on each `@Test` test function?
2. Is it reset on each test cycle?

but, `Connection reset on each test run.` seems to provide a better understanding to users.

junoyoon: "run" is official word in grinder world. See.. "run count"

junoyoon: The reason why I recommend "Connection reset on each test run" rather than"Connection reset after one test cycle", it's little bit shorter than what you used.

imbyungjun (approved):

songeunwoo (approved):

PR #781, opened 2021-03-12
Fix partial read response body

imbyungjun:
- Support partial read response body in HTTP1 (HTTP2 is not supported currently)
- Determine response body length as given read bytes, and ignore remaining bytes.
- After partial read, the connection that was used is not reusable.
   The connection has remaining bytes that did not consumed, so next request will be failed with throwing invalid protocol exception (unexpected response caused by remaining bytes).
- If a user set the connection reset option as false, performance test will be stopped by error.


junoyoon:
Could you explain how it works face to face?

donggyu04 (approved):

songeunwoo (approved):

PR #786, opened 2021-04-04
Change deprecated babel-eslint to new module

donggyu04:
Currently, even if there is no settings change, eslint make below error in all `.vue` files.
Because [babel-eslint](https://www.npmjs.com/package/babel-eslint) we are currently using has been deprecated and has not been updated in the past 1 year.
So, I change eslint module to  [@babel/eslint-parser](https://www.npmjs.com/package/@babel/eslint-parser) that has been released for the replacement of [babel-eslint](https://www.npmjs.com/package/babel-eslint) to resolve errors from old modules.


donggyu04:
Move to #787

PR #789, opened 2021-04-08
Change perftest script template to use new HTTP client

imbyungjun:
* Change the default groovy script template to use new HTTP client 
  * Groovy : use a new HTTP client
  * Groovy(Legacy) : use a old HTTP Client
  * Jython : use a old HTTP client (there are no changes)
  * Groovy Gradle Project : use a new HTTP Client


ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyExScriptHandler.java
```diff
@@ -0,0 +1,60 @@
...
+ * @since 3.5.5
+ */
+@Component
+public class GroovyExScriptHandler extends GroovyScriptHandler {
+
+	/**
+	 * Constructor.
+	 */
+	public GroovyExScriptHandler() {
+		this("groovy_ex", "groovy", "Groovy (Experimental)", "groovy", true);
```
junoyoon, line 30: Not anymore experimental.
Please switch groovy to new script template.
and provide GroovyWithOldHttpClientHandler alternatively.

ngrinder-runtime/src/main/java/org/ngrinder/http/HTTPResponse.java
```diff
@@ -38,7 +41,7 @@
 		this.message = message;
 	}
 
-	public byte[] getBodyBytes() {
+	public byte[] getBody() {
```
junoyoon, line 44: Why is this changed?

imbyungjun: I think the suffix `bytes` is unnecessary. When I suddenly watch this method name again, the word `bodyBytes` seems to be unnatural.

donggyu04 (dismissed):

songeunwoo (dismissed):

ngrinder-runtime/src/main/java/org/ngrinder/http/HTTPResponse.java
junoyoon: Sure.. usually I think develops may expect getBody returns Body instance not bytes[]

imbyungjun: Got it. The way to make this more intuitive could be to add a Body. I'll reset it.

donggyu04 (approved):

PR #817, opened 2021-08-03
Change thread shutdown exception log level

donggyu04:
In Jython script, when the running test is terminated by shutdown signal `ShutdownException` is logged into agent logging message with error level. so users seem to be confused by that log. (many users ask me 'What is this error log?')
to reduce confusion, change logging level from error to info.

> I think `ShutdownException` from the worker thread is normal termination not error.

- Before

- After


imbyungjun (approved):
LGTM. It looks like critical error occurred but, it's kind of normal case when the test shutdown in jython.
This change will calm down the users who run the jython test for the first time.

songeunwoo (approved):
I think it's a very good try :)
In addition to this, it would be nice to leave a comment in the logic of which part was modified.

junoyoon (approved):

PR #823, opened 2021-09-02
fix agent run fail

lingsoul:
fix agent run fail


lingsoul:
run run_agent.sh
ERROR starter: Was passed main parameter '$@' but no main parameter was defined

PR #852, opened 2021-12-10
Increase scm column size

donggyu04:
```
com.mysql.cj.jdbc.exceptions.MysqlDataTruncation: Data truncation: Data too long for column 'scm' at row 1
```

Some github scripts have too long names to save them.


imbyungjun (dismissed):

imbyungjun:
image
Add config name validation to make it short than 40

PR #887, opened 2022-07-12
Pass targetHosts property to agent

outstanding1301:
There's no way to get `perfTest.targetHosts` in test script.

I found there's `ngrinder.etc.hosts` property but it indicates `targetHosts` only in specific condition

So I made a feature that pass `targetHosts` property to `Agent` following steps bellow

1. Put `ngrinder.target.hosts` property to `GrinderProperties` in PerfTestService if `perfTest` has `targetHosts`.  
2. Add it to agent property as `targetHosts` in `PropertyBuilder`

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
+++ b/ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
@@ -36,6 +36,7 @@ import static java.util.Collections.singletonList;
 import static javax.net.ssl.SSLSocketFactory.getDefault;
 import static org.ngrinder.common.constants.GrinderConstants.GRINDER_PROP_CONNECTION_RESET;
 import static org.ngrinder.common.constants.GrinderConstants.GRINDER_SECURITY_LEVEL_LIGHT;
+import static org.ngrinder.common.constants.GrinderConstants.GRINDER_PROP_TARGET_HOSTS;
 import static org.ngrinder.common.util.NoOp.noOp;
 import static org.ngrinder.common.util.Preconditions.checkNotEmpty;
 import static org.ngrinder.common.util.Preconditions.checkNotNull;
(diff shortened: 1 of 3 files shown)


donggyu04:
@outstanding1301 

Why do you need targetHosts in your script?

junoyoon:
By the way, this feature seems good.

@imbyungjun review and merge it.

junoyoon (approved):

ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
```diff
@@ -190,6 +191,7 @@ public String buildJVMArgumentWithoutMemory() {
 		}
 
 		addParam(jvmArguments, properties.getProperty("grinder.param", ""));
+		addTargetHosts(jvmArguments);
```
imbyungjun, line 194: What this method does is identical with what the `addCustomDns()` method does.
You can access target hosts with `System.getProperty("ngrinder.etc.hosts")`.
It is not necessary to add this attribute.

imbyungjun:
@outstanding1301 

What's the difference between `ngrinder.etc.hosts` and `targetHosts`? I don't understand this property addition is necessary.

imbyungjun:
These are the result what I printing `ngrinder.etc.hosts` and `targetHosts`.

**Perftest Configuration**
image

**Test Script**
```groovy
grinder.logger.info("{}", System.getProperty("ngrinder.etc.hosts"))
grinder.logger.info("{}", System.getProperty("targetHosts"))
```

**Result Log**
```
2022-09-28 14:56:35,859 INFO  mylaptop.local:127.0.0.1,localhost:127.0.0.1,www.naver.com:0.0.0.1,www.google.com:1.1.1.1
2022-09-28 14:56:35,859 INFO  www.naver.com:0.0.0.1,www.google.com:1.1.1.1
```

Issue #885, opened 2022-07-07
실행시 no port is available


BGHyeon:
#### 오류 log
<pre><code>
[...]

PR #756, opened 2021-02-04
Look up controller ip from dns when agent re-tries to connect to controller

donggyu04:
To support dynamic IP to a specific domain, when the agent re-reties to connect to a disconnected controller  It looks up controller IP from DNS.

> TTL properties refered to [AWS developer guide](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html)

PR #894, opened 2022-08-16
Add InetAddress reachable check

imbyungjun:
resolved: https://github.com/naver/ngrinder/issues/885

On macOS, `InetAddress.getLocalHost()` returns weired address and it derives no port available exception. 
Available port check is executed with the wrong IP address, so the controller can not acquire any ports.

---

After digging a few days, It comes out not an application bug, it is a kind of PC settings issue.
No major logic changes in this PR. I just make it clearer how to deal with this issue.

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
+++ b/ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
@@ -52,6 +52,7 @@ import static org.ngrinder.common.util.NoOp.noOp;
 @SuppressWarnings("SameParameterValue")
 public abstract class NetworkUtils {
 	private static final Logger LOGGER = LoggerFactory.getLogger(NetworkUtils.class);
+	private static final int MAX_REACHABLE_TIMEOUT = 10;
 
 	/**
 	 * Get the local host address, try to get actual IP.
@@ -143,8 +144,7 @@ public abstract class NetworkUtils {
 		return true;
 	}
 
-	static InetAddress getFirstNonLoopbackAddress(boolean preferIpv4, boolean preferIPv6)
-		throws SocketException {
+	static InetAddress getFirstNonLoopbackAddress(boolean preferIpv4, boolean preferIPv6) throws SocketException {
 		Enumeration<?> en = getNetworkInterfaces();
 		while (en.hasMoreElements()) {
 			NetworkInterface i = (NetworkInterface) en.nextElement();
@@ -223,7 +223,13 @@ public abstract class NetworkUtils {
 		if (StringUtils.isNotBlank(ip)) {
 			try {
 				inetAddress = InetAddress.getByName(ip);
-			} catch (Exception e) {
+				if (!inetAddress.isReachable(MAX_REACHABLE_TIMEOUT)) {
+					processException(
+						"Can not check available ports because given local IP address '" + ip + "' is unreachable. " +
+							"Please check the `/etc/hosts` file or manually specify the local IP address in `${NGRINDER_HOME}/system.conf`."
+					);
+				}
+			} catch (SecurityException | IOException e) {
 				noOp();
 			}
 		}
@@ -504,8 +510,7 @@ public abstract class NetworkUtils {
 			Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
 			while (interfaces.hasMoreElements()) {
 				NetworkInterface iface = interfaces.nextElement();
-				if (iface.isLoopback() || !iface.isUp())
-					continue;
+				if (iface.isLoopback() || !iface.isUp()) continue;
 
 				Enumeration<InetAddress> addresses = iface.getInetAddresses();
 				while (addresses.hasMoreElements()) {
--- a/ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
+++ b/ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
@@ -52,9 +52,6 @@ import static org.ngrinder.common.util.NoOp.noOp;
 @SuppressWarnings("SameParameterValue")
 public abstract class NetworkUtils {
 	private static final Logger LOGGER = LoggerFactory.getLogger(NetworkUtils.class);
-	public static String DEFAULT_LOCAL_HOST_ADDRESS = getLocalHostAddress();
-	public static String DEFAULT_LOCAL_HOST_NAME = getLocalHostName();
-	public static List<InetAddress> DEFAULT_LOCAL_ADDRESSES = getAllLocalNonLoopbackAddresses(false);
 
 	/**
 	 * Get the local host address, try to get actual IP.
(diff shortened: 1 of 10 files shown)


ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
```diff
@@ -64,7 +64,19 @@ public abstract class NetworkUtils {
...
+				while (addresses.hasMoreElements()) {
+					InetAddress address = addresses.nextElement();
+					if (address instanceof Inet4Address && !address.isLoopbackAddress()) {
+						localHost = address;
+						break;
+					}
+				}
+			} else {
+				localHost = InetAddress.getLocalHost();
+			}
```
junoyoon, line 79: I think if someone use VPN, what happen?
Where does this logic com from?

ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
```diff
@@ -52,9 +52,6 @@
 @SuppressWarnings("SameParameterValue")
 public abstract class NetworkUtils {
 	private static final Logger LOGGER = LoggerFactory.getLogger(NetworkUtils.class);
-	public static String DEFAULT_LOCAL_HOST_ADDRESS = getLocalHostAddress();
-	public static String DEFAULT_LOCAL_HOST_NAME = getLocalHostName();
-	public static List<InetAddress> DEFAULT_LOCAL_ADDRESSES = getAllLocalNonLoopbackAddresses(false);
```
imbyungjun, line 57: After trying some cases, it doesn't seem like something that should be handled on application level. I leave a comment on  https://github.com/naver/ngrinder/issues/885#issuecomment-1225240338.

After changing DNS configuration, it could derive anther [issue](https://stackoverflow.com/questions/33289695/inetaddress-getlocalhost-slow-to-run-30-seconds) that may slow down the nGrinder but, it is a kind of configuration issue also. 
Spring framework also recognizes the issue and it prints a warning log on startup like below.
```
2022-08-24 15:09:07,248 WARN  StartupInfoLogger.java:116 : InetAddress.getLocalHost().getHostName() took 5013 milliseconds to respond. Please verify your network configuration (macOS machines may need to add entries to /etc/hosts).
```

How about removing only the use of NetworkUtils constants from this PR?
Or how about just closing this PR?

ngrinder-core/src/main/java/net/grinder/AgentController.java
```diff
@@ -144,18 +143,15 @@ public void run() {
 							connector = m_connectorFactory.create(agentConfig.getConnectionAgentPort());
 							occupyConnectionAgentSocket();
 						} else {
-							String controllerIP = getIP(agentConfig.getControllerHost());
-							agentConfig.setControllerIP(controllerIP);
-							connector = m_connectorFactory.create(controllerIP, agentConfig.getControllerPort());
+							connector = m_connectorFactory.create(agentConfig.getControllerIP(), agentConfig.getControllerPort());
```
imbyungjun, line 146: @donggyu04 

While removing the NetworkUtils constants, I found out this code block unnecessary.
This line changed in https://github.com/naver/ngrinder/pull/756. But, before entering to `AgentController#run`, `NGrinderAgentStarter` set up the controller IP address like below.
https://github.com/naver/ngrinder/blob/345f1d37bf380c5e788dbd28aa3508dd980a36bd/ngrinder-core/src/main/java/org/ngrinder/NGrinderAgentStarter.java#L137-L139

So, I think looking up the DNS again here is an unnecessary code block. Is it ok to remove these? 

Should I remove those lines on `NGrinderAgentStarter`? (If I remove those code lines on `NGrinderAgentStarter`, `AgentController` still check the DNS again.)

donggyu04: @imbyungjun 

`NGrinderAgentStarter`'s code is called only once at agent startup.
but, this code exists for the case below

1. Agent connect to Controller(`www.mycontroller.com:10.10.10.10`) successfully
2. Controller shut down & restart with changed ip with same domain (`www.mycontroller.com:10.10.22.22`)
3. If remove this code block, agent will still retry connection to Controller with old ip(10.10.10.10) that saved when it startup

imbyungjun: Thx. I'll remove `NGrinderAgentStarter`'s controller IP setup code.

ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
```diff
@@ -223,7 +223,13 @@ public static List<Integer> getAvailablePorts(String ip, int size, int from, int
 		if (StringUtils.isNotBlank(ip)) {
 			try {
 				inetAddress = InetAddress.getByName(ip);
-			} catch (Exception e) {
+				if (!inetAddress.isReachable(MAX_REACHABLE_TIMEOUT)) {
+					processException(
+						"Can not check available ports because given local IP address '" + ip + "' is unreachable. " +
+							"Please check the `/etc/hosts` file or manually specify the local IP address in `${NGRINDER_HOME}/system.conf`."
+					);
+				}
```
imbyungjun, line 231: The issue #885 was hard to understand because the error message was not appropriate.
Adding reachability check and giving a detail error message will make it clearer.

donggyu04 (approved):

ngrinder-core/src/main/java/net/grinder/util/NetworkUtils.java
```diff
@@ -226,7 +223,13 @@ public static List<Integer> getAvailablePorts(String ip, int size, int from, int
 		if (StringUtils.isNotBlank(ip)) {
 			try {
 				inetAddress = InetAddress.getByName(ip);
-			} catch (Exception e) {
+				if (!inetAddress.isReachable(MAX_REACHABLE_TIMEOUT)) {
+					processException(
+						"Can not check available ports because given local IP address '" + ip + "' is unreachable. " +
```
junoyoon, line 228: IP address. <-- . is missing.

imbyungjun: I was intended to make a message like `IP address '${ip}' is unreachable.`. 
In this case, exception message is printed like this.
```
Caused by: org.ngrinder.common.exception.NGrinderRuntimeException: Can not check available ports because given local IP address '123.234.123.234' is unreachable. Please check the `/etc/hosts` file or manually specify the local IP address in `${NGRINDER_HOME}/system.conf`.
	at org.ngrinder.common.util.ExceptionUtils.processException(ExceptionUtils.java:38)
	at net.grinder.util.NetworkUtils.getAvailablePorts(NetworkUtils.java:227)
	at org.ngrinder.perftest.service.ConsoleManager.init(ConsoleManager.java:76)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        ...
```

Issue #895, opened 2022-08-17
ngrinder-fronted module build failed. (ES-check)


umyonghyun:
### 1. Describe the bug 🐞 
:exclamation: DO NOT ASK QUESTION HERE PLEASE USE [DISCUSSION](https://github.com/naver/ngrinder/discussions) INSTEAD.

i pull ngrinder source code for [this issue](https://github.com/naver/ngrinder/issues/893).
but when i build this code in develop branch, error has occured. 
what shoud i do?

### 2. Reproduction steps
1. github source pull
2. Intellij > build > build Artifacts > .....war

### 3. Environment
- Controller
     - OS:  Mac Monterey (12.1 MacBook Pro(16형, 2021년 모델))
     - Browser: X
     - JDK version: coretto-11.0.14
     - Controller version: develop branch version
- Agent
     - OS:
     - JDK version:
     - Agent version:

### 4. Screenshots

> Task :ngrinder-frontend:checkES5
> debug: ES-Check: Going to check files using version 5
> debug: ES-Check: checking ../ngrinder-controller/build/classes/main/static/js/app.js
> debug: ES-Check: failed to parse file: ../ngrinder-controller/build/classes/main/static/js/app.js 
> error: SyntaxError: Assigning to rvalue (77:24424)
> error: ES-Check: there were 1 ES version matching errors.

```
          ES-Check Error:
          ----
          · erroring file: ../ngrinder-controller/build/classes/main/static/js/app.js
ES-Check: there were 1 ES version matching errors.

          · error: SyntaxError: Assigning to rvalue (77:24424)
          · see the printed err.stack below for context
          ----

          SyntaxError: Assigning to rvalue (77:24424)
    at Parser.pp$4.raise (~/.npm/_npx/b7764aab07c01327/node_modules/acorn/dist/acorn.js:3459:15)
    at Parser.pp$9.checkPatternErrors (~/.npm/_npx/b7764aab07c01327/node_modules/acorn/dist/acorn.js:777:29)
    at Parser.pp$7.toAssignable (~/.npm/_npx/b7764aab07c01327/node_modules/acorn/dist/acorn.js:1962:47)
    at Parser.pp$5.parseMaybeAssign (~/.npm/_npx/b7764aab07c01327/node_modules/acorn/dist/acorn.js:2484:23)
    at Parser.pp$8.parseVar (~/.npm/_npx/b7764aab07c01327/node_modules/acorn/dist/acorn.js:1303:26)
[...]

PR #896, opened 2022-08-24
Remove es-check gradle task

imbyungjun:
Resolved: #895

Main role of the `es-check` task was to detect IE compatibility issues.
Since the IE has retired, remove the `es-check` task. 
We expect that the most of modern web browers are compatible.


donggyu04 (approved):

junoyoon (approved):
good

PR #918, opened 2022-10-18
fix: fileType autodetect error when miss extension, it should be UNKNOWN

colin-lee:
change fileType detect error when resource fileName without extension. Eg. resources/cookies, will not copy to agent


junoyoon:
I think you shoud not merge commit from develop branch. instead, please rebase it.

junoyoon:
I close this PR becaz it remain as uncompleted.
By the way, I understood the issue so @imbyungjun will make a new PR for this.

PR #922, opened 2022-10-19
Fix gradle command execution on Oshi 6.x

imbyungjun:
Recently, nGrinder bumps Oshi from `5.2.2` to `6.1.6`.
And Oshi changes the `runNative()` method to pass environment variables.
`runNative()` method is a wrapper of `Runtime.getRuntime().exec()` that captures the output in a list.

```java
public static List<String> runNative(String[] cmdToRunWithArgs) {
    return runNative(cmdToRunWithArgs, DEFAULT_ENV); // DEFAULT_ENV == "LC_ALL=C"
}
```

This change makes executing gradle goes wrong. So, I changed to pass nothing as environment variable.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
@@ -87,7 +87,7 @@ public class GroovyGradleProjectScriptHandler extends GroovyProjectScriptHandler
 		if (results.isEmpty()) {
 			return false;
 		}
-		return results.stream().noneMatch(str -> str.contains("FAILED"));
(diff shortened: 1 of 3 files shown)


ngrinder-controller/src/main/java/org/ngrinder/script/handler/GroovyGradleProjectScriptHandler.java
```diff
@@ -87,7 +87,7 @@ protected boolean isSuccess(List<String> results) {
 		if (results.isEmpty()) {
 			return false;
 		}
-		return results.stream().noneMatch(str -> str.contains("FAILED"));
+		return results.stream().noneMatch(str -> str.contains("FAILED") || str.contains("ERROR"));
```
donggyu04, line 90: 👍

donggyu04 (approved):

PR #973, opened 2023-07-21
Fix to use controller.host property for ip resolve

junoyoon:
in the system.conf, controller.host field exists but is not used. 
(Instead old property (controller.ip) was used..)
Fix this....


junoyoon:
wrong pr.. close.

PR #974, opened 2023-07-21
Use controller.host property instead of old controller.ip property

junoyoon:
controller.ip property was used even after this property has changed its name to controller.host.

Fix this

### DIFF
--- a/ngrinder-controller/src/main/resources/controller-properties.map
+++ b/ngrinder-controller/src/main/resources/controller-properties.map
@@ -17,7 +17,7 @@ controller.monitor_port,13243,monitor.listen.port
 controller.url,,ngrinder.http.url,http.url
 controller.console_port_base,12000,ngrinder.console.portbase
 controller.controller_port,16001,ngrinder.agent.control.port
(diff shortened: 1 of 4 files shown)


junoyoon:
@donggyu04 @imbyungjun  please check this. and merge if ok.

donggyu04 (approved):

ngrinder-controller/src/main/resources/controller-properties.map
```diff
@@ -17,7 +17,7 @@ controller.monitor_port,13243,monitor.listen.port
 controller.url,,ngrinder.http.url,http.url
 controller.console_port_base,12000,ngrinder.console.portbase
 controller.controller_port,16001,ngrinder.agent.control.port
-controller.ip,,ngrinder.controller.ipaddress,ngrinder.controller.ip
+controller.host,,controller.ip,ngrinder.controller.ipaddress,ngrinder.controller.ip
```
donggyu04, line 20: oh in `system.conf`, `controller.host` was being used since 2014.

imbyungjun (approved):

PR #993, opened 2024-01-03
Add access control for webhook

imbyungjun:


ngrinder-controller/src/main/java/org/ngrinder/infra/webhook/controller/WebhookApiController.java
```diff
@@ -47,7 +47,7 @@ public class WebhookApiController {
 	private final WebhookActivationService webhookActivationService;
 
 	@PostMapping
-	public void save(@RequestBody WebhookConfig webhookConfig) {
+	public void save(User user, @RequestBody WebhookConfig webhookConfig) {
```
donggyu04, line 50: It needs to be modified `user.id` for creating new webhook instead of [WebhookConfig.creatorId](https://github.com/naver/ngrinder/blob/develop/ngrinder-controller/src/main/java/org/ngrinder/infra/webhook/model/WebhookConfig.java#L49)

imbyungjun: changed to replace creatorId as authenticated userId. 4c49e3b6ac9b09b68f7b85db65020d21341b8574

donggyu04 (approved):

PR #995, opened 2024-01-08
Change yaml parser to yamlbeans

imbyungjun:
snakeyaml 1.x has security vulnerability. So, I tried to bump it up to 2.x. But, the snakeyaml is related to spring framework also. When the snakeyaml is found on classpath, spring try to use the given snakeyaml to handling yaml property file.

Most of spring-boot versions are depend on snakeyaml 1.x except latest version of spring-boot. I think that upgrade spring-boot to latest version will take long time. So, just change the yaml parser to [yamlbeans](https://github.com/EsotericSoftware/yamlbeans) that has most stars library in [yaml official website](https://yaml.org/) except snakeyaml.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/service/GitHubFileEntryService.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/service/GitHubFileEntryService.java
@@ -1,5 +1,6 @@
 package org.ngrinder.script.service;
 
+import com.esotericsoftware.yamlbeans.YamlReader;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
@@ -26,7 +27,6 @@ import org.tmatesoft.svn.core.SVNException;
 import org.tmatesoft.svn.core.SVNURL;
 import org.tmatesoft.svn.core.auth.BasicAuthenticationManager;
 import org.tmatesoft.svn.core.wc.*;
-import org.yaml.snakeyaml.Yaml;
 
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


ngrinder-controller/src/main/java/org/ngrinder/script/service/GitHubFileEntryService.java
```diff
@@ -278,25 +278,26 @@ public Set<GitHubConfig> getAllGitHubConfig(User user) throws FileNotFoundExcept
 		return getAllGithubConfig(gitConfigYaml);
 	}
 
-	private Set<GitHubConfig> getAllGithubConfig(FileEntry gitConfigYaml) {
+	protected Set<GitHubConfig> getAllGithubConfig(FileEntry gitConfigYaml) {
```
donggyu04, line 281: Why modifier was changed?

donggyu04 (dismissed):

ngrinder-controller/src/main/java/org/ngrinder/script/service/GitHubFileEntryService.java
imbyungjun: I was changed it for adding a unit test. But, it doesn't need to change access modifier.
I will change it back to private and add a unit test to verify that successfully parse the `.gitconfig.yml` file.

ngrinder-controller/src/test/java/org/ngrinder/script/service/GitHubFileEntryServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Autowired
+    private GitHubFileEntryService gitHubFileEntryService;
+
+    @Autowired
+    private MockFileEntityRepository mockFileEntityRepository;
+
+    @Before
+    public void before() throws IOException {
+        File file = new File(System.getProperty("java.io.tmpdir"), "repo");
```
donggyu04, line 35: How about use [Junit TempDir](https://www.baeldung.com/junit-5-temporary-directory#2-tempdir-on-an-instance-field)?
It will be deleted automatically after test

donggyu04 (approved):

ngrinder-controller/src/test/java/org/ngrinder/script/service/GitHubFileEntryServiceTest.java
imbyungjun: We are using junit 4.13.1 😭. And TempDir seems to be supported since Junit5.
https://github.com/naver/ngrinder/blob/f6d067271ad10716fe4f75ba77216a9fc7143306/gradle.properties#L4

PR #996, opened 2024-01-09
Fix unit test

imbyungjun:
fixed some unit tests

### DIFF
--- a/ngrinder-core/src/test/java/net/grinder/AgentControllerTest.java
+++ b/ngrinder-core/src/test/java/net/grinder/AgentControllerTest.java
@@ -174,7 +174,7 @@ public class AgentControllerTest extends AbstractMultiGrinderTestBase {
 
 		FileUtils.copyFileToDirectory(
 			new File(this.getClass().getResource("/long-time-prepare-test.py").getFile()),
-			new File("./tmp/agent-home/tmp_1/file-store/_default/incoming")
+			new File("./tmp/agent-home/tmp_0/file-store/_default/incoming")
 		);
 
 		URL scriptUrl = this.getClass().getResource("/long-time-prepare-test.properties");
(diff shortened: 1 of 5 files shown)


ngrinder-core/src/test/java/org/ngrinder/AbstractMultiGrinderTestBase.java
```diff
@@ -43,13 +43,13 @@ abstract public class AbstractMultiGrinderTestBase {
 
 	@Before
 	public void agentInit() {
-		agentConfig1 = new AgentConfig.NullAgentConfig(1);
+		agentConfig1 = new AgentConfig.NullAgentConfig(0);
```
imbyungjun, line 46: I think that setting `ngrinder.agent.home` property as `./tmp/agent-home` could derive resource contention. I wanted to fix it but no time. I'll change it if there is a next time.

ngrinder-core/src/test/java/net/grinder/AgentControllerTest.java
```diff
@@ -172,25 +174,22 @@ public void testConsoleCommunicationSettingTimeout() throws GrinderProperties.Pe
...
-				System.out.println("The stop signal is received " + stopReason);
-				if (stopReason.equals(SCRIPT_ERROR)) {
-					timeouted.setValue(true);
-				}
-			}
-		});
+		console1.addListener(stopReason -> {
+            // Notice: it couldn't distinguish between a script error or
+            // timed out of the keepalive connection.
+            System.out.println("The stop signal is received " + stopReason);
```
donggyu04, line 188: How about use `private static final Logger LOGGER = LoggerFactory.getLogger(AgentControllerTest.class);`?

donggyu04 (approved):

PR #998, opened 2024-01-16
Fix failing to call mvn or gradle at MAVEN_HOME, GRADLE_HOME respectively in Windows (Resolves #975)

### LINKED ISSUE #975: Can't run Maven and Gradle By NGrinder Controller
### 1. Describe the bug 🐞 
Can't find Maven and Gradle when running NGrinder Controller even though maven_home and gradle_home are set already. 
( I think it happens only on windows. Reason will be described below )

It seems this problem caused by InstallationChecker.java file. ( ngrinder-controller/src/main/java/org/ngrinder/starter/InstallationChecker.java) 

first and second line of enum, there's command only "mvn -version", "gradle -version".
Unfortunately, on windows, that doens't work. Those only works on linux and might be mac.
On windows platform, that command should be changed to "mvn.cmd -version", "gradle.bat -version"

I hope it will help this project.

### 2. Reproduction steps
1. Go to directory Ngrinder controller war file exists.
2. run war file using java command "java -Djava.io.tmpdir=%NGRINDER_HOME%\lib -jar [warfilename].war
3. See error : 
"WARNING: Maven isn't installed, You can't run Maven groovy scripts. Please install Maven and set MAVEN_HOME."
"WARNING: Gradle isn't installed, You can't run Gradle groovy scripts. Please install Gradle and set GRADLE_HOME."

### 3. Environment
- Controller
     - OS: Windows NT
     - Browser: none
     - JDK version: 14
     - Controller version: 3.5.8

### 4. Screenshots
If necessary, add screenshots to help explain your problem.
bug_report_ngrinder

facewise:
There were two major factors causing this bug.

1. Calling `System.getenv()` on initializing enums was failed on Windows.
2. Getting the standard output of the process using `oshi.util.ExecutingCommand.runNative()` was failed on Windows due to Linux-specific command scripts.

### Changes
1. Changed the time calling `System.getenv()` after initializing.
2. Set OS specific command scripts checking installation using `com.sun.jna.Platform.isWindows()`.

Since `com.sun.jna.Platform` sets the static values of current OS on start, no needed to call `System.getProperty("os.name")` again.
But considered that the dependency including `com.sun.jna.Platform` may be removed in future, it could be the best option to check by calling `System.getProperty("os.name")`.

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/starter/InstallationChecker.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/starter/InstallationChecker.java
@@ -20,6 +20,9 @@
  */
 package org.ngrinder.starter;
 
+import com.sun.jna.Platform;
+
+import java.io.File;
 import java.util.List;
 
 import static java.lang.System.*;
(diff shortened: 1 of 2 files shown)


imbyungjun (approved):

imbyungjun:
@facewise 

When I install maven and gradle in Windows env, there are executable binary files each.

maven
- mvn
- mvn.cmd

gradle
- gradle
- gradle.bat

So, the original check command that using `mvn` and `gradle` works fine for me.
I don't understand why the check command for Windows needs to be changed.
Actually, I don't use Windows that much. Let me know if there is a reason the check command should be change.

imbyungjun:
@facewise 

Thanks, The issue only occurred when running the command using oshi.

I have successfully reproduced the issue and confirmed that it is resolved in your PR.
Thanks again for your contribution to nGrinder.

PR #1000, opened 2024-02-05
Add object input filter on Connector

imbyungjun:
And bump target JDK version up to 11 also.

### DIFF
--- a/ngrinder-core/src/main/java/net/grinder/communication/Connector.java
+++ b/ngrinder-core/src/main/java/net/grinder/communication/Connector.java
@@ -23,6 +23,7 @@ package net.grinder.communication;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputFilter;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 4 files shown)


ngrinder-core/src/main/java/net/grinder/communication/Connector.java
```diff
@@ -197,6 +200,23 @@ static ConnectDetails read(InputStream in) throws CommunicationException {
...
 			final ObjectInputStream objectInputStream = new ObjectInputStream(in);
+			objectInputStream.setObjectInputFilter(filterInfo -> {
+				if (filterInfo.serialClass() == null) {
+					return ObjectInputFilter.Status.UNDECIDED;
+				}
+
+				if (Address.class.isAssignableFrom(filterInfo.serialClass())
+					|| ConnectionType.class.isAssignableFrom(filterInfo.serialClass())
+					|| Enum.class.isAssignableFrom(filterInfo.serialClass())
+					|| filterInfo.serialClass().getCanonicalName().startsWith("net.grinder.engine")
```
donggyu04, line 211: Can be attacked if class name is `net.grinder.engine.Attacker.java`?

Issue #940, opened 2022-11-20
why V3.5.7 Script Validation Error？


cwk0099:
os: amazon linux
How to run ngrinder：docker
Steps to reproduce the bug：always when click validate button

response in /script/api/validate
{
  "message" : "Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: Could not initialize class org.ngrinder.common.util.SystemInfoUtils",
  "success" : false,
  "stackTrace" : "org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: Could not initialize class org.ngrinder.common.util.SystemInfoUtils\n\tat org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)\n\tat org.ngrinder.security.PluggablePreAuthFilter.doFilter(PluggablePreAuthFilter.java:52)\n\tat org.ngrinder.infra.servlet.PluggableServletFilter.doFilter(PluggableServletFilter.java:61)\n\tat org.ngrinder.security.PluggablePreAuthFilter.doFilter(PluggablePreAuthFilter.java:52)\n\tat org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1589)\n\tat org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)\n\tat org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)\nCaused by: java.lang.NoClassDefFoundError: Could not initialize class org.ngrinder.common.util.SystemInfoUtils\n\tat net.grinder.engine.agent.PropertyBuilder.buildJVMArgumentWithoutMemory(PropertyBuilder.java:208)\n\tat net.grinder.engine.agent.LocalScriptTestDriveService.doValidate(LocalScriptTestDriveService.java:131)\n\tat org.ngrinder.script.service.ScriptValidationService.validate(ScriptValidationService.java:117)\n\tat org.ngrinder.script.controller.FileEntryApiController.validate(FileEntryApiController.java:392)\n\t... 7 more\n"
}

joonseolee:
I'm going to say me either. 
I run both of the controller and agent in the amazon linux docker.  
It didn't work after pushing the validate button in the script tab.  
fortunately, 3.5.5-p1 version is working now.  
I would like to keep an eye on this issue.

AIMWLI:
[...]

PR #1004, opened 2024-02-27
Fix oshi

imbyungjun:
resolve https://github.com/naver/ngrinder/issues/940

This issue seems like ohsi error that is occurred in docker container env only.
Currently, base JDK version is changed to 11. So, updated oshi version to work with JDK 11. 
And expect this change will fix the oshi issue also.


ngrinder-core/src/main/java/net/grinder/engine/agent/PropertyBuilder.java
```diff
@@ -205,7 +204,7 @@ public String buildJVMArgumentWithoutMemory() {
 			addAdditionalJavaOpt(jvmArguments);
 		}
 
-		String jdkVersion = getJDKVersion();
+		String jdkVersion = System.getProperty("java.specification.version");
```
imbyungjun, line 207: `getJDKVersion()` method called only here. And it is not related with Oshi unlike the other methods. So, remove the method and change to check system property here directly.

imbyungjun: This change also make Script Validation function doesn't depends on SystemInfoUtils. So, I expect that NoClassDefFoundError does not occur anymore.

donggyu04 (approved):
👍

PR #1017, opened 2024-05-07
Add authorization settings for API

imbyungjun:


ngrinder-frontend/src/js/components/perftest/list/List.vue
```diff
@@ -420,7 +420,7 @@
                         }
 
                         this.tests[target.index].status = updatedStatus;
-                        this.runningSummary = `${res.data.perfTestInfo.length} ${this.i18n('perfTest.list.runningSummary')}`;
+                        this.runningSummary = `${res.data.runningTestsCount} ${this.i18n('perfTest.list.runningSummary')}`;
```
imbyungjun, line 423: It unnecessarily exposed other users' performance test information. So, changed it.

ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
```diff
@@ -65,6 +66,7 @@
 @RestController
 @RequestMapping("/script/api")
 @RequiredArgsConstructor
+@PreAuthorize("hasAnyRole('A', 'S', 'U')")
```
imbyungjun, line 69: All of FileEntiry related APIs need authorization even if the method doesn't have a User parameter.

donggyu04 (approved):

PR #1018, opened 2024-05-14
Change to use snakeyaml with SafeConstructor

imbyungjun:
Revert commit to use snakeyaml again.
And change to use SafeConstructor.


ngrinder-controller/src/main/java/org/ngrinder/script/service/GitHubFileEntryService.java
```diff
@@ -280,24 +281,23 @@ public Set<GitHubConfig> getAllGitHubConfig(User user) throws FileNotFoundExcept
...
-				GitHubConfig config = objectMapper.convertValue(gitConfigMap, GitHubConfig.class);
-
-				if (gitHubConfig.contains(config)) {
-					throw new InvalidGitHubConfigurationException("GitHub configuration '"
-							+ config.getName() + "' is duplicated.\nPlease check your .gitconfig.yml");
-				}
-
-				gitHubConfig.add(config);
+		// Yaml is not thread safe. so create it every time.
+		Yaml yaml = new Yaml(new SafeConstructor());
```
imbyungjun, line 285: Actually, the one thing that changed is here. Use the SafeConstructor.

PR #1022, opened 2024-06-13
Fix some tests

imbyungjun:


ngrinder-controller/src/main/java/org/ngrinder/infra/init/ClassPathInit.java
```diff
@@ -46,6 +46,9 @@ public void init() {
 		final String systemClasspath = System.getProperty("java.class.path", StringUtils.EMPTY);
 		for (String pathEntry : systemClasspath.split(File.pathSeparator)) {
 			final File parentFile = ObjectUtils.defaultIfNull(new File(pathEntry).getParentFile(), new File("."));
+			if (!parentFile.isDirectory()) {
+				continue;
+			}
```
imbyungjun, line 51: I'm not sure if this class is needed at this time. Just filtered out to fix test context loading failures.

PR #1038, opened 2025-03-25
Fix: XSS Vulnerability in File Upload Descriptions

hsuk04:
### **Description**
This PR addresses a cross-site scripting vulnerability by implementing XssPreventer.escape() on all description parameters during file uploads. 
[Fix ](https://github.com/naver/ngrinder/commit/2e09d20fda1edbae3d166f408f0d2666439403f1)was done previously in a separate controller file but not applied here.

### **References**
Related fix [commit 2e09d20](https://github.com/naver/ngrinder/commit/2e09d20fda1edbae3d166f408f0d2666439403f1).

### DIFF
--- a/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
+++ b/ngrinder-controller/src/main/java/org/ngrinder/script/controller/FileEntryApiController.java
@@ -49,6 +49,7 @@ import org.springframework.http.ContentDisposition;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
+import com.nhncorp.lucy.security.xss.XssPreventer;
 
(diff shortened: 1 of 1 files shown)


imbyungjun:
Thank you so much for your PR. I really appreciate your contribution!

I wanted to ask if you have checked whether it is possible to exploit an XSS vulnerability by inputting values into the fileEntry description. 
Currently, nGrinder’s Front-end has been migrated to Vue, so there is no need to use XssPreventer for escaping XSS vulnerabilities.

I tested the scenario you mentioned as shown below, and I confirmed that no vulnerability exists.
image

However, if you have found a case where an XSS vulnerability can be exploited, please let me know.

Once again, thank you for your effort and contribution.

hsuk04:
Hi, @imbyungjun.
Thanks for the clarification, it was probably an oversight on my end and not vulnerable to XSS.
Will be closing the PR.

